PR #7, opened 2020-11-05
[linagora/lgs/linshare/products/linshare-mobile-apps#544] Linter

hoangdat:


hoangdat:
@manhduc-nguyen please review this PR

manhduc-nguyen:
@hoangdat your last commit failed from CI test, can you check that pls.

pubspec.yaml
```diff
@@ -75,6 +75,7 @@ dependencies:
   # The following adds the Cupertino Icons font to your application.
   # Use with the CupertinoIcons class for iOS style icons.
   cupertino_icons: ^1.0.0
+  pedantic: ^1.9.2
```
manhduc-nguyen, line 78: I think we should use a specific version for this library `1.9.2`, not a dynamic version `^1.9.2`. `^` means that when you run `pub get` it might install a newer version than `1.9.2` if available. I think this is not good because we have CI for this, with newer `pedantic` library version it might fail CI job at anytime

data/lib/src/datasource_impl/document_datasource_impl.dart
```diff
@@ -47,7 +47,7 @@ class DocumentDataSourceImpl implements DocumentDataSource {
 
   @override
   Future<FileUploadState> upload(FileInfo fileInfo, Token token, Uri baseUrl) async {
-    File file = File(fileInfo.filePath + fileInfo.fileName);
+    var file = File(fileInfo.filePath + fileInfo.fileName);
```
manhduc-nguyen, line 50: I think it can be `final` instead of `var`. same to others in this commit.

data/test/authentication_datasource_test.dart
```diff
@@ -0,0 +1,138 @@
...
+
+    setUp(() {
+      _deviceManager = MockDeviceManager();
+      _linShareHttpClient = MockLinShareHttpClient();
+      _authenticationDataSource = AuthenticationDataSource(_linShareHttpClient, _deviceManager);
+    });
+
+    test('createPermanentToken should create permanentToken success', () async {
+      when(_deviceManager.getDeviceUUID())
+        .thenAnswer((realInvocation) async => Future.value('12345-bde44'));
```
manhduc-nguyen, line 56: since you are not using param `realInvocation `, so you can write it like this `.thenAnswer((_) async =>`. same to others.

manhduc-nguyen (changes requested):

pubspec.yaml
hoangdat: Got it. But I like that failed, we need to update each time we failed

data/lib/src/datasource_impl/document_datasource_impl.dart
hoangdat: ok

data/test/authentication_datasource_test.dart
hoangdat: got it

manhduc-nguyen:
LGTM!

manhduc-nguyen (approved):
LGTM!

manhduc-nguyen (approved):
LGTM!

tnstung (approved):

PR #24, opened 2020-11-24
LA-583 Responsive My Space

imGok:
1st Breakpoint < 640
image

2nd Breakpoint >= 640
Screenshot_1606313164

3rd Breakpoint - > 960
image

### DIFF
--- a/lib/presentation/util/helper/responsive_widget.dart
+++ b/lib/presentation/util/helper/responsive_widget.dart
@@ -1,0 +1,77 @@
+// LinShare is an open source filesharing software, part of the LinPKI software
+// suite, developed by Linagora.
+//
+// Copyright (C) 2020 LINAGORA
+//
+// This program is free software: you can redistribute it and/or modify it under the
+// terms of the GNU Affero General Public License as published by the Free Software
+// Foundation, either version 3 of the License, or (at your option) any later version,
+// provided you comply with the Additional Terms applicable for LinShare software by
+// Linagora pursuant to Section 7 of the GNU Affero General Public License,
+// subsections (b), (c), and (e), pursuant to which you must notably (i) retain the
+// display in the interface of the “LinShare™” trademark/logo, the "Libre & Free" mention,
+// the words “You are using the Free and Open Source version of LinShare™, powered by
+// Linagora © 2009–2020. Contribute to Linshare R&D by subscribing to an Enterprise
+// offer!”. You must also retain the latter notice in all asynchronous messages such as
+// e-mails sent with the Program, (ii) retain all hypertext links between LinShare and
+// http://www.linshare.org, between linagora.com and Linagora, and (iii) refrain from
+// infringing Linagora intellectual property rights over its trademarks and commercial
+// brands. Other Additional Terms apply, see
+// <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf>
+// for more details.
+// This program is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for
+// more details.
+// You should have received a copy of the GNU Affero General Public License and its
+// applicable Additional Terms for LinShare along with this program. If not, see
+// <http://www.gnu.org/licenses/> for the GNU Affero General Public License version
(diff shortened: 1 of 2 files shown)


imGok:
Before changes : 
tabletDemoOld

hoangdat:
Bigger button is quite not good.

lib/presentation/widget/myspace/my_space_widget.dart
```diff
@@ -88,7 +89,7 @@ class _MySpaceWidgetState extends State<MySpaceWidget> {
                     converter: (store) => store.state.mySpaceState.viewState,
                     builder: (context, viewState) {
                       return viewState.fold(
-                          (failure) => Container(),
+                          (failure) => SizedBox.shrink(),
```
imGok, line 92: Replaced `Container()` by `SizedBox.shrink()` for empty areas

=> https://stackoverflow.com/questions/53455358/how-to-present-an-empty-view-in-flutter

lib/presentation/widget/myspace/my_space_widget.dart
```diff
@@ -163,46 +164,50 @@ class _MySpaceWidgetState extends State<MySpaceWidget> {
...
-                    ),
-                  )
-                : Container()
+            _buildModifiedDocumentText(AppLocalizations.of(context).item_last_modified(
+                  document.modificationDate.getMMMddyyyyFormatString())),
+            _buildSharedIcon(document.isShared())
           ],
         ),
-      ),
+      ) : null,
```
imGok, line 210: null and not `SizedBox.shrink()` because ListTile auto align the content

imGok:
IMO It is necessary to enlarge the elements because it's too little on tablet / Android TV

lib/presentation/widget/myspace/my_space_widget.dart
manhduc-nguyen: nice

manhduc-nguyen:
have you tried to change screen to `landscape mode` ?

lib/presentation/util/helper/responsive_widget.dart
```diff
@@ -0,0 +1,78 @@
...
+import 'dart:ui';
+
+import 'package:flutter/material.dart';
+
+class ResponsiveWidget extends StatelessWidget {
+  final Widget largeScreen;
+  final Widget mediumScreen;
+  final Widget smallScreen;
+
+  static final double widthDevice = window.physicalSize.width;
```
hoangdat, line 41: What is the comparison with `MediaQuery`?

lib/presentation/util/helper/responsive_widget.dart
```diff
@@ -0,0 +1,78 @@
...
+
+  static final int minLargeWidth = 2048;
+  static final int minMediumWidth = 1536;
```
hoangdat, line 44: IMO, this pixel is not realistic for mobile and tablet device, please explain about unit using for width and height. I want to make clear about `logicalPixel` and `physicalPixel`

hoangdat: Base on https://material.io/design/layout/pixel-density.html#pixel-density.

I understand that, your width is in `physicalPixel`. It will be different in other device with other density. So I think, we should not depend on `physicalPixel`. We need to depend on `logicalPixel`. Please investigate more. Thanks

hoangdat: Note: we can access logical pixel via: `MediaQuery.of(context).size`

lib/presentation/util/helper/responsive_widget.dart
imGok: As you said, it gives the `physicalPixel` which is bad

lib/presentation/util/helper/responsive_widget.dart
imGok: After talking with designer, it seems I misunderstood the device size. You're right, using logicalPixel is way more effective and sustainable. I put breakpoints in description, this is working fine for me tell me what you think :)

hoangdat (approved):

imGok:
> have you tried to change screen to `landscape mode` ?

landscape in 1st breakpoint become 2nd breakpoint
image

landscape in 2nd breakpoint become 3rd breakpoint
image

manhduc-nguyen (approved):

hoangdat (approved):

PR #26, opened 2020-11-25
set enableFlutterDriverExtension() function just work with profile mode

Trongk58d:


hoangdat:
Please create an `ADR` to explain about that problem. What is the context, reason, ...?

Please ask @manhduc-nguyen to get the support :+1:

hoangdat (approved):

PR #32, opened 2020-12-07
Fix the message icon and content that CI job pushes to Twake

tnstung:
This is to fix the message icon and the message content that pushes to the FT Mobile channel once a build job is done.


hoangdat (approved):

hoangdat:
You can merge whenever it is ok for you @tnstung

PR #55, opened 2020-12-22
LA-614 Add Log to Interactors

imGok:


hoangdat:
This logging option only visible on debug? Not visible on release? @imGok

imGok:
> This logging option only visible on debug? Not visible on release? @imGok

still investigating on that because I didn't confirmed. Can't run in release mode on my PC atm

hoangdat:
> > This logging option only visible on debug? Not visible on release? @imGok
> 
> still investigating on that because I didn't confirmed. Can't run in release mode on my PC atm

how about with `flutter run release`

imGok:
> how about with `flutter run release`

yes I've tried this with simulator and it didn't work will try with real device

hoangdat:
Please make your solution compare with `fimber`: https://pub.dev/packages/fimber and other librarys

PR #57, opened 2020-12-23
Building path with path's parameters suggestion

ngohado:
Sorry. I tried to implement building path with parameters without know we have already exist :( . But I still push and create PR to show you another view to create path with path's parameters.

### DIFF
--- a/data/lib/data.dart
+++ b/data/lib/data.dart
@@ -31,6 +31,7 @@ export 'src/datasource_impl/autocomplete_datasource_impl.dart';
 
 // util
 export 'src/util/device_manager.dart';
+export 'src/network/config/endpoint.dart';
 
 // model
 export 'src/network/model/response/document_response.dart';
--- a/data/lib/src/network/config/endpoint.dart
+++ b/data/lib/src/network/config/endpoint.dart
@@ -35,7 +35,7 @@ import 'package:domain/domain.dart';
 class Endpoint {
   static final String rootPath = '/linshare/webservice/rest/user/v2';
   static final String download = '/download';
-  static final ServicePath authentication = ServicePath('/jwt');
+  static final ServicePath authentication = ServicePath(r'/jwt/$tokenId');
 
   static final ServicePath authorizedUser = ServicePath('/authentication/authorized');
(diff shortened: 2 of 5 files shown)


data/lib/src/network/config/endpoint.dart
```diff
@@ -44,7 +44,7 @@ class Endpoint {
 
   static final ServicePath sharedSpaces = ServicePath('/shared_spaces');
 
-  static final ServicePath autocomplete = ServicePath('/autocomplete');
+  static final ServicePath autocomplete = ServicePath(r'/autocomplete/$query');
```
hoangdat, line 47: can you explain more clear? what is `r`

data/lib/src/network/linshare_http_client.dart
```diff
@@ -119,7 +121,7 @@ class LinShareHttpClient {
   Future<List<AutoCompleteResult>> getSharingAutoComplete(
       AutoCompletePattern autoCompletePattern) async {
     final List resultJson = await _dioClient.get(Endpoint.autocomplete
-        .withPathParameter(autoCompletePattern.value)
+        .withPathParameters([PathParameter('query', autoCompletePattern.value)])
```
hoangdat, line 124: cool. I like it

data/test/network/endpoint_ext_test.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:domain/domain.dart';
+import 'package:data/src/network/config/endpoint.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('endpoint extension tests', () {
+    test('Service path build with path parameters', () {
```
hoangdat, line 7: please test more cases, `null`, `exception`, ...

data/lib/data.dart
```diff
@@ -31,6 +31,7 @@ export 'src/datasource_impl/autocomplete_datasource_impl.dart';
 
 // util
 export 'src/util/device_manager.dart';
+export 'src/network/config/endpoint.dart';
```
manhduc-nguyen, line 34: I don't think we need to export `endpoint.dart `

PR #77, opened 2021-01-04
LA-64 Share multiple files

imGok:
demo_multiple_selection_final

### DIFF
--- a/lib/l10n/intl_en.arb
+++ b/lib/l10n/intl_en.arb
@@ -248,11 +248,18 @@
     "type": "text",
     "placeholders": {}
   },
-  "items": "{numberOfItems} item(s)",
+  "items": "{numberOfItems,plural, =0{No item}=1{{numberOfItems} item}other{{numberOfItems} items}}",
   "@items": {
     "type": "text",
     "placeholders": {
       "numberOfItems": {}
     }
   },
+  "items_selected": "{numberOfItems,plural, =1{{numberOfItems} item selected}other{{numberOfItems} items selected}}",
+  "@items_selected": {
+    "type": "text",
+    "placeholders": {
+      "numberOfItems": {}
+    }
+  }
 }
--- a/lib/l10n/intl_en.arb
+++ b/lib/l10n/intl_en.arb
@@ -237,5 +237,22 @@
   "@browse": {
     "type": "text",
     "placeholders": {}
-  }
+  },
+  "select_all": "Select all",
+  "@select_all": {
+    "type": "text",
+    "placeholders": {}
+  },
+  "unselect_all": "Unselect all",
+  "@unselect_all": {
+    "type": "text",
+    "placeholders": {}
+  },
+  "items": "{numberOfItems} item(s)",
+  "@items": {
+    "type": "text",
+    "placeholders": {
+      "numberOfItems": {}
+    }
+  },
 }
--- a/lib/l10n/intl_fr.arb
+++ b/lib/l10n/intl_fr.arb
@@ -248,11 +248,18 @@
     "type": "text",
     "placeholders": {}
   },
-  "items": "{numberOfItems} item(s)",
+  "items": "{numberOfItems,plural, =0{No item}=1{{numberOfItems} item}other{{numberOfItems} items}}",
   "@items": {
     "type": "text",
     "placeholders": {
       "numberOfItems": {}
     }
   },
+  "items_selected": "{numberOfItems,plural, =1{{numberOfItems} item selected}other{{numberOfItems} items selected}}",
+  "@items_selected": {
+    "type": "text",
+    "placeholders": {
+      "numberOfItems": {}
+    }
+  }
 }
--- a/lib/l10n/intl_fr.arb
+++ b/lib/l10n/intl_fr.arb
@@ -237,5 +237,22 @@
   "@browse": {
     "type": "text",
     "placeholders": {}
-  }
+  },
+  "select_all": "Select all",
+  "@select_all": {
+    "type": "text",
+    "placeholders": {}
+  },
+  "unselect_all": "Unselect all",
+  "@unselect_all": {
+    "type": "text",
+    "placeholders": {}
+  },
+  "items": "{numberOfItems} item(s)",
+  "@items": {
+    "type": "text",
+    "placeholders": {
+      "numberOfItems": {}
+    }
+  },
 }
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -42,6 +42,7 @@ import 'package:flutter_driver/driver_extension.dart';
 import 'package:flutter/foundation.dart';
 
 import 'presentation/redux/states/app_state.dart';
+import 'presentation/util/extensions/color_extension.dart';
 import 'presentation/widget/initialize/initialize_widget.dart';
 
 void main() async {
@@ -82,6 +83,7 @@ class LinShareApp extends StatelessWidget {
           // the app on. For desktop platforms, the controls will be smaller and
           // closer together (more dense) than on mobile platforms.
           visualDensity: VisualDensity.adaptivePlatformDensity,
+          unselectedWidgetColor: AppColor.unselectedElementColor
         ),
         home: FutureBuilder(
             future: getIt.allReady(),
--- a/lib/presentation/model/file/selected_presentation_file.dart
+++ b/lib/presentation/model/file/selected_presentation_file.dart
@@ -17,7 +17,8 @@
 // http://www.linshare.org, between linagora.com and Linagora, and (iii) refrain from
 // infringing Linagora intellectual property rights over its trademarks and commercial
 // brands. Other Additional Terms apply, see
-// <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf>
+// <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf
+//
 // for more details.
 // This program is distributed in the hope that it will be useful, but WITHOUT ANY
 // WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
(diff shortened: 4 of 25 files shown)


lib/presentation/redux/states/my_space_state.dart
```diff
@@ -39,30 +41,55 @@ import 'package:linshare_flutter_app/presentation/redux/states/linshare_state.da
...
 
   @override
   LinShareState sendViewState({Either<Failure, Success> viewState}) {
-    return MySpaceState(viewState, documentList);
+    return MySpaceState(viewState, documentList, isSelectingDocuments);
   }
 
   LinShareState setDocuments({Either<Failure, Success> viewState, List<Document> newDocumentList}) {
-    return MySpaceState(viewState, newDocumentList);
+    return MySpaceState(viewState, { for (var document in newDocumentList) document: false}, isSelectingDocuments);
```
hoangdat, line 68: so, all the selected document will be unselected? 
what is better for this case?

lib/presentation/redux/states/my_space_state.dart
```diff
@@ -39,30 +41,55 @@ import 'package:linshare_flutter_app/presentation/redux/states/linshare_state.da
 
 @immutable
 class MySpaceState extends LinShareState {
-  final List<Document> documentList;
+  final Map<Document, bool> documentList;
+  final bool isSelectingDocuments;
```
hoangdat, line 45: How about strong typing the state with an enum?
```dart
enum SelectMode { ACTIVE, INACTIVE }
```

hoangdat (commented):
Overview quite good. please move on.

lib/main.dart
```diff
@@ -82,6 +83,7 @@ class LinShareApp extends StatelessWidget {
           // the app on. For desktop platforms, the controls will be smaller and
           // closer together (more dense) than on mobile platforms.
           visualDensity: VisualDensity.adaptivePlatformDensity,
+          unselectedWidgetColor: AppColor.unselectedElementColor
```
manhduc-nguyen, line 86: Sr, can you share a bit about this? I don't understand what it is used for.

lib/presentation/redux/states/my_space_state.dart
imGok: all list will be unselected when the user is fetching new documents, I think it's a better case avoiding selected documents not existing anymore

lib/presentation/redux/states/my_space_state.dart
imGok: Sure why not I'll give it a try

lib/main.dart
imGok: This will colors all unselected widget like checkbox, toggle, radio etc 

> The color used for widgets in their inactive (but enabled) state. For example, an unchecked checkbox. Usually contrasted with the accentColor. See also disabledColor.

In Figma, this is the color of all these items

lib/presentation/redux/states/my_space_state.dart
hoangdat: @dieptran88 Can you give me an UX for it? Should we disable select mode on reload request?

dieptran88: i think we should not auto unselect when refreshing page, bc :

lib/presentation/redux/states/my_space_state.dart
```diff
@@ -39,30 +41,55 @@ import 'package:linshare_flutter_app/presentation/redux/states/linshare_state.da
 
 @immutable
 class MySpaceState extends LinShareState {
-  final List<Document> documentList;
+  final Map<Document, bool> documentList;
```
ngohado, line 44: About data structure concept, I think we use `Map` here is not appropriate. We use `List` when we need listing something, special for the things need to be in order, and `Map` for input and output. 

I think `Document` list may sometimes need to be ordered by date, name,... And I think we use `List` is more appropriate than `Map`

ngohado: I think we can consider to create a wrapper class included two properties is document, and select status. We also need to consider abstract it. It could be:
```
class SelectableElement<T> {
    T element;
    SelectMode selectMode;
}
```
We can reuse it with any type of element

lib/presentation/redux/states/my_space_state.dart
imGok: Done it, it looks cleaner thx

lib/presentation/redux/states/my_space_state.dart
imGok: Love ur idea, check again please :)

lib/main.dart
hoangdat: cool

lib/presentation/redux/states/my_space_state.dart
```diff
@@ -35,34 +35,64 @@ import 'package:domain/domain.dart';
...
-    return MySpaceState(viewState, newDocumentList);
+    final selectedElements = documentList.where((element) => element.selectMode == SelectMode.ACTIVE).map((element) => element.element).toList();
+
+    return MySpaceState(viewState,
+      {for (var document in newDocumentList)
+          if (selectedElements.contains(document))
+            SelectableElement<Document>(document, SelectMode.ACTIVE)
+          else SelectableElement<Document>(document, SelectMode.INACTIVE)}.toList(),
+      selectMode);
+  }
```
hoangdat, line 75: Please check the performance of the case: `refresh during selected mode`?. I concerned about when the list is rebuild.

lib/presentation/widget/myspace/my_space_viewmodel.dart
```diff
@@ -231,8 +232,25 @@ class MySpaceViewModel extends BaseViewModel {
...
+    }
+  }
+
+  void cancelSelection() {
+    store.dispatch(MySpaceClearSelectedDocumentsAction());
+  }
+
   @override
   void onDisposed() {
+    cancelSelection();
```
hoangdat, line 253: please make the GIF for the case: 
- Select some documents in MySpace
- Open Side menu -> SharedSpace
- Back to MySpace: by `Side Menu` or `Back press button on Android`

imGok: test_dat

lib/presentation/redux/states/my_space_state.dart
imGok: It doesn't bug or anything on Nexus API 30, have you seen smth ?

imGok:
Hi @dieptran88, after `sharing` the user returns in My Space with same selected elements, is that a good UX ? (see demo in description)

lib/presentation/redux/states/my_space_state.dart
hoangdat: No^^. So good, I will check in real device

hoangdat:
> Hi @dieptran88, after `sharing` the user returns in My Space with same selected elements, is that a good UX ? (see demo in description)

oh, no, it is not good for the UX

lib/presentation/widget/myspace/my_space_widget.dart
```diff
@@ -81,6 +84,41 @@ class _MySpaceWidgetState extends State<MySpaceWidget> {
...
+                  StoreConnector<AppState, MySpaceState>(
+                    converter: (Store<AppState> store) => store.state.mySpaceState,
+                    builder: (context, state) {
+                      return state.selectMode == SelectMode.ACTIVE ? ListTile(
+                        leading: SvgPicture.asset(
+                          imagePath.icSelectAll,
+                          width: 28,
+                          height: 28,
+                          fit: BoxFit.fill,
+                          color: state.documentList.every((value) => value.selectMode == SelectMode.ACTIVE) ?
```
manhduc-nguyen, line 96: since you're using this in many places, please create an extension for documentList to detect the list has active mode item or not.

lib/l10n/intl_en.arb
```diff
@@ -255,4 +255,11 @@
       "numberOfItems": {}
     }
   },
+  "items_selected": "{numberOfItems} items selected",
```
hoangdat, line 258: small note: `Intl` support plurals. Please check in details: https://pub.dev/packages/intl

lib/presentation/model/file/selected_presentation_file.dart
```diff
@@ -0,0 +1,47 @@
...
+// for the GNU Affero General Public License version
+//
+// 3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf
+// for
+//
+// the Additional Terms applicable to LinShare software.
+
+import 'package:equatable/equatable.dart';
+
+class SelectedPresentationFile extends Equatable {
```
manhduc-nguyen, line 39: I checked design, I think we also need `mediaType`, for displaying file type icon.

lib/presentation/widget/myspace/my_space_viewmodel.dart
hoangdat: Sound good for the back press. How about:
- Select some documents in MySpace
- Open Side menu -> SharedSpace
- Open Side menu -> MySpace

lib/presentation/model/file/selected_presentation_file.dart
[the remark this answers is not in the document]
hoangdat: please ignore it temporary. I created a ticket to handle it later: #87

hoangdat: We will do it later: #87

lib/l10n/intl_fr.arb
```diff
@@ -237,5 +237,22 @@
   "@browse": {
     "type": "text",
     "placeholders": {}
-  }
+  },
+  "select_all": "Select all",
```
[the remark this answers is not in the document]
imGok: Done!

lib/presentation/widget/myspace/my_space_viewmodel.dart
imGok: my_space_to_shared_space_multiple_selection

lib/presentation/widget/myspace/my_space_widget.dart
imGok: Sure great idea

lib/l10n/intl_en.arb
imGok: Done thx

lib/presentation/widget/myspace/my_space_viewmodel.dart
hoangdat: so good.

hoangdat (approved):

lib/presentation/widget/myspace/my_space_widget.dart
```diff
@@ -137,12 +193,12 @@ class _MySpaceWidgetState extends State<MySpaceWidget> {
                       height: 120,
                       fit: BoxFit.fill))
                     .text(AppLocalizations.of(context).common_error_occured_message)
-                    .build() : _buildMySpaceListView(context, mySpaceState.documentList)),
+                    .build() : _buildMySpaceListView(context, mySpaceState.documentList.map((element) => element.element).toList())),
```
ngohado, line 196: I prefers use `mySpaceState.documentList` `List<SelectableElement<Document>>` instead of convert to `List<Document>`. 

We can use immediately `SelectMode` instead of searching every item build by `isDocumentSelected ()`. It takes a lot of computation although it's unnecessary :(

lib/presentation/widget/myspace/my_space_widget.dart
```diff
@@ -208,14 +264,27 @@ class _MySpaceWidgetState extends State<MySpaceWidget> {
...
-          height: 24,
-          fit: BoxFit.fill,
-      ), onPressed: () => mySpaceViewModel
-          .openContextMenu(context, document, contextMenuActionTiles(context, document))),
+      trailing: StoreConnector<AppState, MySpaceState>(
+        converter: (store) => store.state.mySpaceState,
+        builder: (context, state) {
+          return state.selectMode == SelectMode.ACTIVE ?
+            Checkbox(
+              value: state.isDocumentSelected(document),
```
ngohado, line 272: I means at above comment about use `mySpaceState.documentList` `List<SelectableElement<Document>>`

lib/presentation/widget/myspace/my_space_widget.dart
```diff
@@ -208,14 +264,27 @@ class _MySpaceWidgetState extends State<MySpaceWidget> {
...
       ) : null,
-      trailing: IconButton(
-        icon: SvgPicture.asset(
-          imagePath.icContextMenu,
-          width: 24,
-          height: 24,
-          fit: BoxFit.fill,
-      ), onPressed: () => mySpaceViewModel
-          .openContextMenu(context, document, contextMenuActionTiles(context, document))),
+      trailing: StoreConnector<AppState, MySpaceState>(
```
ngohado, line 267: I prefer use `StoreConnector` to rebuild `ListView` instead of element of each item. Because, when we works with 10000.. items or more. It's create a lot of connector to Store. And it's seem terrible '_'

Just my guest base on my knowledge. Plz help me test it. About use `StoreConnector` to rebuild `ListView`, indeed we just rebuild items are displaying on screen, I think the performance is good ^^

lib/presentation/widget/myspace/my_space_widget.dart
imGok: Yes it's not necessary to map the list you're right 

> We can use immediately SelectMode instead of searching every item build by isDocumentSelected ()

Couldn't get ur point can you be more specific please ?

imGok: Nvm I get it, nice comment

lib/presentation/widget/myspace/my_space_widget.dart
imGok: I think it's better to put StoreConnector to lowest elements and make the viewModel really convenient since it will be rebuilt everytime this viewModel changed.
In my case, it would be better to have 
```
      trailing: StoreConnector<AppState, SelectMode>(
        converter: (store) => store.state.mySpaceState.selectMode,
        builder: (context, selectMode) {
```
However you're right, I put a storeConnector just for the select mode I could have just give this in parameter. 

About performance, I don't think it will be a problem since we don't have a lot of documents lists at the same time

lib/presentation/widget/myspace/my_space_widget.dart
imGok: u ok with that ? 5e8bc8c76416792688dba828109994cf46e4a99f

ngohado (approved):

hoangdat (approved):

PR #78, opened 2021-01-05
LA-59: Destination picker

manhduc-nguyen:
Screen Shot 2021-01-07 at 17 41 36.  Screen Shot 2021-01-07 at 17 41 48

### DIFF
--- a/lib/presentation/widget/destination_picker/destination_picker_widget.dart
+++ b/lib/presentation/widget/destination_picker/destination_picker_widget.dart
@@ -1,0 +1,290 @@
+// LinShare is an open source filesharing software, part of the LinPKI software
+// suite, developed by Linagora.
+//
+// Copyright (C) 2021 LINAGORA
+//
+// This program is free software: you can redistribute it and/or modify it under the
+// terms of the GNU Affero General Public License as published by the Free Software
+// Foundation, either version 3 of the License, or (at your option) any later version,
+// provided you comply with the Additional Terms applicable for LinShare software by
+// Linagora pursuant to Section 7 of the GNU Affero General Public License,
+// subsections (b), (c), and (e), pursuant to which you must notably (i) retain the
+// display in the interface of the “LinShare™” trademark/logo, the "Libre & Free" mention,
+// the words “You are using the Free and Open Source version of LinShare™, powered by
+// Linagora © 2009–2020. Contribute to Linshare R&D by subscribing to an Enterprise
+// offer!”. You must also retain the latter notice in all asynchronous messages such as
+// e-mails sent with the Program, (ii) retain all hypertext links between LinShare and
+// http://www.linshare.org, between linagora.com and Linagora, and (iii) refrain from
+// infringing Linagora intellectual property rights over its trademarks and commercial
+// brands. Other Additional Terms apply, see
+// <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf>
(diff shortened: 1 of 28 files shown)


lib/presentation/widget/shared_space/file_surfing/workgroup_detail_files_widget.dart
```diff
@@ -64,14 +69,16 @@ class _WorkGroupDetailFilesWidgetState extends State<WorkGroupDetailFilesWidget>
...
+          nodeSurfingType: widget.nodeSurfingType,
         ),
         Align(
           alignment: Alignment.bottomCenter,
           child: SizedBox(
             height: 76,
             child: Stack(
               children: [
-                Align(
+                widget.nodeSurfingType == NodeSurfingType.normal
```
ngohado, line 80: I think we should seperate whole widget, it is more clear and updateable in future in case we have more type (just my opinion)
```
switch (type) {
    case NodeSurfingType.normal:
        return _buildDocumentNormalItem();
    case NodeSurfingType. destinationPicker:
        return _buildDestinationPickerItem();
}
```

lib/presentation/widget/destination_picker/destination_picker_widget.dart
```diff
@@ -0,0 +1,274 @@
...
+    return Positioned(
+        bottom: 0.0,
+        left: 20.0,
+        right: 20.0,
+        child: SizedBox(
+          height: 56.0,
+          child: Row(
+            textDirection: TextDirection.rtl,
+            children: [
+              TextButton(
```
manhduc-nguyen, line 180: Please skip dummy code, I''ll change this later.

manhduc-nguyen:
@hoangdat @imGok @ngohado Please review this guys.

lib/presentation/widget/destination_picker/destination_picker_widget.dart
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  Widget _buildSharedSpaceName(String sharedSpaceName) {
+    return Text(
+      sharedSpaceName,
+      maxLines: 1,
+      style: TextStyle(fontSize: 14, color: AppColor.documentNameItemTextColor),
+    );
+  }
+}
```
hoangdat, line 274: we can not reuse any of this from SharedSpace?

lib/main.dart
```diff
@@ -65,6 +65,7 @@ class LinShareApp extends StatelessWidget {
     return StoreProvider(
       store: store,
       child: MaterialApp(
+        debugShowCheckedModeBanner: false,
```
imGok, line 68: Why do you disable this ?

manhduc-nguyen: oh mb

imGok: No worries, just asking because It doesn't bother IMO

manhduc-nguyen: we have other ways to disable this in local only, so I should remove this.

lib/presentation/widget/destination_picker/destination_picker_widget.dart
manhduc-nguyen: Then we need to create some widget builders to reuse this, but I don't think this is necessary in this case :(

manhduc-nguyen:
@hoangdat @imGok please check my fixup commit

hoangdat (approved):

PR #131, opened 2021-01-21
LA-48 Delete Multiple Work Group Nodes Presentation Layer

imGok:
demo_delete_file_multiple


imGok:
@hoangdat @manhduc-nguyen please review guys, tmr I'll look for an other solution for upload button because it's not convenient like this

manhduc-nguyen (approved):
LGTM

hoangdat (approved):

PR #152, opened 2021-01-25
LA-135: add network state connection observer

manhduc-nguyen:
Screen Shot 2021-01-25 at 17 25 35

### DIFF
--- a/lib/presentation/redux/states/network_connectivity_state.dart
+++ b/lib/presentation/redux/states/network_connectivity_state.dart
@@ -36,9 +36,10 @@ import 'package:domain/src/state/failure.dart';
 import 'package:domain/src/state/success.dart';
 import 'package:flutter/foundation.dart';
 import 'package:linshare_flutter_app/presentation/redux/states/linshare_state.dart';
+import 'package:equatable/equatable.dart';
 
 @immutable
-class NetworkConnectivityState extends LinShareState {
+class NetworkConnectivityState extends LinShareState with EquatableMixin {
   final ConnectivityResult connectivityResult;
 
   NetworkConnectivityState(Either<Failure, Success> viewState, this.connectivityResult) : super(viewState);
@@ -65,4 +66,10 @@ class NetworkConnectivityState extends LinShareState {
   NetworkConnectivityState startLoadingState() {
     return NetworkConnectivityState(Right(LoadingState()), connectivityResult);
   }
+
+  @override
+  List<Object> get props => [
+    ...super.props,
+    connectivityResult
+  ];
 }
--- a/lib/presentation/redux/states/network_connectivity_state.dart
+++ b/lib/presentation/redux/states/network_connectivity_state.dart
@@ -1,7 +1,7 @@
 // LinShare is an open source filesharing software, part of the LinPKI software
 // suite, developed by Linagora.
 //
-// Copyright (C) 2020 LINAGORA
+// Copyright (C) 2021 LINAGORA
 //
 // This program is free software: you can redistribute it and/or modify it under the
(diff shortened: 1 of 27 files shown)


lib/presentation/widget/home/home_viewmodel.dart
```diff
@@ -69,6 +74,19 @@ class HomeViewModel extends BaseViewModel {
     });
   }
 
+  void _registerNetworkConnectivityState() async {
+    store.dispatch(_setNetworkStateAction(await Connectivity().checkConnectivity()));
+    _connectivityStreamSubscription = Connectivity().onConnectivityChanged.listen((ConnectivityResult result) {
```
hoangdat, line 79: Please provide `Connectivity` as a singleton.

lib/presentation/redux/states/network_connectivity_state.dart
```diff
@@ -0,0 +1,68 @@
...
+class NetworkConnectivityState extends LinShareState {
+  final ConnectivityResult connectivityResult;
+
+  NetworkConnectivityState(Either<Failure, Success> viewState, this.connectivityResult) : super(viewState);
+
+  factory NetworkConnectivityState.initial() {
+    return NetworkConnectivityState(Right(IdleState()), null);
+  }
+
+  LinShareState setNetworkState(ConnectivityResult connectivityResult) {
```
imGok, line 50: Methods in state should return ``NetworkConnectivityState`` imo

lib/presentation/redux/states/network_connectivity_state.dart
```diff
@@ -0,0 +1,68 @@
...
+import 'package:connectivity/connectivity.dart';
+import 'package:dartz/dartz.dart';
+import 'package:domain/src/state/failure.dart';
+import 'package:domain/src/state/success.dart';
+import 'package:flutter/foundation.dart';
+import 'package:linshare_flutter_app/presentation/redux/states/linshare_state.dart';
+
+@immutable
+class NetworkConnectivityState extends LinShareState {
+  final ConnectivityResult connectivityResult;
```
imGok, line 42: I don't think creating a state just for that is good why not put it in UI state ?

lib/presentation/widget/home/home_viewmodel.dart
```diff
@@ -69,6 +76,19 @@ class HomeViewModel extends BaseViewModel {
...
+  void _registerNetworkConnectivityState() async {
+    store.dispatch(_setNetworkStateAction(await _connectivity.checkConnectivity()));
+    _connectivityStreamSubscription = _connectivity.onConnectivityChanged.listen((ConnectivityResult result) {
+      store.dispatch(_setNetworkStateAction(result));
+    });
+  }
+
+  ThunkAction<AppState> _setNetworkStateAction(ConnectivityResult connectivityResult) {
+    return (Store<AppState> store) async {
+      store.dispatch(SetNetworkConnectivityStateAction(connectivityResult));
```
imGok, line 88: You can directly send this action you don't need a ThunkAction wdyt ?

lib/presentation/redux/states/network_connectivity_state.dart
manhduc-nguyen: Oh no, I think its not really related to `UIState`. We should separate it.

lib/presentation/redux/states/network_connectivity_state.dart
manhduc-nguyen: you are right.

manhduc-nguyen:
@imGok @hoangdat pls check my fix up commit. I also added unit test

imGok (approved):

hoangdat (approved):

PR #164, opened 2021-02-01
LA-140 Verify quota upload My Space

imGok:


domain/lib/src/usecases/quota/verify_quota_my_space_upload_interactor.dart
```diff
@@ -0,0 +1,54 @@
...
+// applicable Additional Terms for LinShare along with this program. If not, see
+// <http://www.gnu.org/licenses/> for the GNU Affero General Public License version
+//  3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf> for
+//  the Additional Terms applicable to LinShare software.
+
+import 'package:dartz/dartz.dart';
+import 'package:domain/domain.dart';
+import 'dart:core';
+
+class VerifyQuotaMySpaceUploadInteractor {
```
hoangdat, line 36: How about verify with multiple upload file @imGok

hoangdat (commented):
quite good to me

domain/lib/src/usecases/quota/verify_quota_my_space_upload_interactor.dart
imGok: Yeah I'll edit it

data/test/fixture/quota_fixture.dart
```diff
@@ -0,0 +1,50 @@
...
+//
+
+import 'package:data/src/network/model/response/account_quota_response.dart';
+import 'package:domain/domain.dart';
+
+final quotaId1 = QuotaId('q1');
+
+final accountQuotaResponse1 = AccountQuotaResponse(
+  quota: QuotaSize(10),
+  usedSpace: QuotaSize(20),
```
manhduc-nguyen, line 40: oh, why usedSpace is greater than quota :( we should create a realistic test data

data/test/fixture/quota_fixture.dart
```diff
@@ -0,0 +1,50 @@
...
+  maxFileSize: QuotaSize(30),
+  maintenance: true
+);
+
+final accountQuota1 = AccountQuota(
+  quota: QuotaSize(10),
```
manhduc-nguyen, line 46: same

domain/lib/src/usecases/quota/verify_quota_my_space_upload_interactor.dart
manhduc-nguyen: About verify multiple file, I think we should get quota info for only 1 time in the interactor, then validate for all files. it will reduce network calls and energy consuming. What do u think? @imGok @hoangdat

hoangdat: Totally agree @manhduc-nguyen

imGok: Yes I agree that's what I was wondering thx for your tip

data/test/fixture/quota_fixture.dart
imGok: Yes MB, I used these only for mocking so I don't really use its info ^^

lib/presentation/widget/home/home_widget.dart
```diff
@@ -162,8 +162,17 @@ class _HomeWidgetState extends State<HomeWidget> {
               if (failure is FilePickerFailure
                   || failure is FileUploadFailure
                   || failure is WorkGroupDocumentUploadFailure) {
-                appToast.showToast(AppLocalizations.of(context).upload_failure_text);
+                appToast.showErrorToast(AppLocalizations.of(context).upload_failure_text);
               }
+
+              if (failure is TooBigFilesMySpaceFailure) {
+                appToast.showErrorToast(AppLocalizations.of(context).tooBigFiles(failure.tooBigFiles.length));
```
manhduc-nguyen, line 169: you should rebase from `master` now, since we handle this in another class

lib/presentation/manager/upload_and_share_file/upload_and_share_file_manager.dart
```diff
@@ -256,4 +264,33 @@ class UploadShareFileManager {
...
+    final accountQuotaResult = (await _getQuotaInteractor.execute(quotaUuid));
+
+    if (accountQuotaResult.isLeft()) {
+      _store.dispatch(UploadFileAction(accountQuotaResult));
+      return false;
+    }
+
+    final totalFilesSize = filesInfos.map((fileInfo) => fileInfo.fileSize).reduce((value, element) => value + element);
+
+    final accountQuota = (accountQuotaResult.getOrElse(() => null) as AccountQuotaViewState).accountQuota;
```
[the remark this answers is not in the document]
imGok: > Moreover, please think that, we don't care to the logic of `_getQuotaInteractor.execute(quotaUuid)`, we only care to the input and output of this, so we need to handle the output exactly.

I don't get your point here can you be more specific pls :D

lib/presentation/manager/upload_and_share_file/upload_and_share_file_manager.dart
```diff
@@ -256,4 +264,33 @@ class UploadShareFileManager {
       genericUsers,
     );
   }
+
+  Future<bool> hasEnoughQuotaAndMaxFileSize({List<FileInfo> filesInfos}) async {
```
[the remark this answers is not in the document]
imGok: Working on it!

lib/presentation/manager/upload_and_share_file/upload_and_share_file_manager.dart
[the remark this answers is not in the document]
hoangdat: - What happen if `accountQuotaResult` is not AccountQuotaViewState?
- What happen if we can not get `accountQuota`?
- Please use the advantage of Either to solve that problem, we should avoid cast without checking type. Your code is too risky. Please try to solve that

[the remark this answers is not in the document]
imGok: Are you ok with this ?

imGok:
@hoangdat @manhduc-nguyen check again pls

lib/presentation/manager/quota/verify_quota_manager.dart
```diff
@@ -61,7 +61,18 @@ class VerifyQuotaManager {
...
+        return success.accountQuota;
+      } else {
+        return null;
+      }
+    });
+
+    if (accountQuota == null) {
+      return false;
+    }
+
```
hoangdat, line 75: it is ok, but we can do it better

hoangdat (approved):

manhduc-nguyen (approved):

Issue #127, opened 2021-01-21
List of files is not auto updated after user deletes one file


hantt12:
### Description
After user successfully deletes a file, list of files should be refreshed and updated automatically

#### Reproduction Steps

1. Connect network and delete a file from MySpace
2. Check the list of files in MySpace

### Expected result
Deleted file is no longer in the list

### Current behavior

Delete file is still there. User has to refresh the list to see it disappear
https://images.zenhubusercontent.com/309616769/9b3c35aa-45b1-4342-9b0c-cb2b62dd737a/img_0571.mp4
### Context

Iphone 11 Promax 
Linshare app 0.1.0 (33)

### Additional information

frequency, log files,...

hoangdat:
@manhduc-nguyen Please find the generic way to update the list after executing operator

manhduc-nguyen:
the problem is after we successfully removed a document, we will dispatch MySpaceAction. then we will dispatch StartMySpaceLoadingAction to start reload the list. StoreConnector only receiving LoadingState and skip RemoveDocumentViewState because it happens too fast.

PR #170, opened 2021-02-02
LA-127: Update file list in MySpace after remove a file

manhduc-nguyen:
Refer to this bug: [https://github.com/linagora/linshare-mobile-flutter-app/issues/127](https://github.com/linagora/linshare-mobile-flutter-app/issues/127)

### DIFF
--- a/lib/presentation/util/toast_message_handler.dart
+++ b/lib/presentation/util/toast_message_handler.dart
@@ -1,0 +1,204 @@
+// LinShare is an open source filesharing software, part of the LinPKI software
+// suite, developed by Linagora.
+//
+// Copyright (C) 2021 LINAGORA
+//
+// This program is free software: you can redistribute it and/or modify it under the
+// terms of the GNU Affero General Public License as published by the Free Software
+// Foundation, either version 3 of the License, or (at your option) any later version,
+// provided you comply with the Additional Terms applicable for LinShare software by
+// Linagora pursuant to Section 7 of the GNU Affero General Public License,
+// subsections (b), (c), and (e), pursuant to which you must notably (i) retain the
+// display in the interface of the “LinShare™” trademark/logo, the "Libre & Free" mention,
+// the words “You are using the Free and Open Source version of LinShare™, powered by
+// Linagora © 2009–2020. Contribute to Linshare R&D by subscribing to an Enterprise
+// offer!”. You must also retain the latter notice in all asynchronous messages such as
+// e-mails sent with the Program, (ii) retain all hypertext links between LinShare and
+// http://www.linshare.org, between linagora.com and Linagora, and (iii) refrain from
+// infringing Linagora intellectual property rights over its trademarks and commercial
+// brands. Other Additional Terms apply, see
+// <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf>
+// for more details.
+// This program is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for
+// more details.
+// You should have received a copy of the GNU Affero General Public License and its
+// applicable Additional Terms for LinShare along with this program. If not, see
+// <http://www.gnu.org/licenses/> for the GNU Affero General Public License version
+//  3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf> for
+//  the Additional Terms applicable to LinShare software.
+//
+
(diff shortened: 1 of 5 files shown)


lib/presentation/util/toast_message_handler.dart
```diff
@@ -0,0 +1,204 @@
...
+      _handleSharedSpaceToastMessage(context, event.sharedSpaceState);
+      _handleNetworkStateToastMessage(context, event.networkConnectivityState);
+    });
+  }
+
+  void _handleMySpaceToastMessage(BuildContext context, MySpaceState mySpaceState) {
+    mySpaceState.viewState.fold((failure) {
+      if (failure is CopyToSharedSpaceFailure || failure is CopyMultipleFilesToSharedSpaceAllFailureViewState) {
+        appToast.showErrorToast(AppLocalizations.of(context).cannot_copy_file_to_shared_space);
+        _cleanMySpaceViewState();
```
hoangdat, line 74: please find the way to move out all function `_cleanMySpaceViewState`. May be using either again

hoangdat: but 1 question: this class is Toast handler, it is good when we clean state here?

hoangdat: how about clean all the state in the obsever of each viewModel?

lib/presentation/widget/myspace/my_space_viewmodel.dart
```diff
@@ -81,7 +84,17 @@ class MySpaceViewModel extends BaseViewModel {
...
+    _storeStreamSubscription = store.onChange.listen((event) {
+      event.mySpaceState.viewState.fold(
+        (failure) => null,
+        (success) {
+          if (success is RemoveDocumentViewState) {
+            getAllDocument();
+          }
+      });
+    });
+  }
```
imGok, line 97: Are you sure it's the only way to fix it ? Why not fetching documents in remove thunk action directly ?

hoangdat: he tried some ways, but it is the best right now. When he did like you said, 2 actions are dispatch to store nearly, so Store connector can not react exactly. `RemoveSuccess` and `Loading` action

lib/presentation/util/toast_message_handler.dart
manhduc-nguyen: agree, clean it in ViewModel seems better

manhduc-nguyen: oh but we have some cases that toast is showed after the ViewModel is destroyed, for example: upload toast message is showed after upload screen is closed, so we can't observe there anymore. So I think we should handle it in `ToastMessageHandler`, WDYT ?

lib/presentation/widget/myspace/my_space_viewmodel.dart
imGok: Ok I see

lib/presentation/util/toast_message_handler.dart
imGok: Since My Space State contains only files list we can reset it when switching views, however account state must not be reset

lib/presentation/widget/myspace/my_space_viewmodel.dart
imGok: This works for me, am I missing smth ?
```
  ThunkAction<AppState> _removeDocumentAction(List<DocumentId> documentIds) {
    return (Store<AppState> store) async {
      await _removeMultipleDocumentsInteractor.execute(documentIds: documentIds)
          .then((result) => result.fold(
              (failure) => store.dispatch(MySpaceAction(Left(failure))),
              (success) => store.dispatch(_removeDocumentSuccessAction(success))));
    };
  }

  ThunkAction<AppState> _removeDocumentSuccessAction(Success success) {
    return (Store<AppState> store) async {
      store.dispatch(MySpaceAction(Right(success)));
      getAllDocument();
    };
  }

  void getAllDocument() {
    store.dispatch(_getAllDocumentAction());
  }
```
ducdelete

manhduc-nguyen: @imGok you are missing the toast message when remove a file successfully

lib/presentation/util/toast_message_handler.dart
manhduc-nguyen: yes, we should reset unnecessary state when jump to other screens

hoangdat: ok with @manhduc-nguyen solution, we will improve it in the future

hoangdat (approved):

PR #173, opened 2021-02-04
Fix signing config for android

hoangdat:
@tnstung Please use this to test deploying APK


hoangdat:
error in apksigner of bitrise. Close this PR because of fix in bitrise.

Issue #174, opened 2021-02-05
UI display incorrectly when upload from outside app


manhduc-nguyen:
![IMG_0083.PNG](https://images.zenhubusercontent.com/5fe5620376fe254973593288/04696d09-d8f3-40b0-972e-429ef0712b60)
### Description
Upload screen UI should display correctly when upload from outside app
#### Reproduction Steps
- Go to photo gallery 
- Pick at least 2 photos
- Select share and select LinShare
- Choose a upload destination is a folder from shared space
- try to open or collapse the file list
### Expected result
- Ui display correctly
### Current behavior
- UI display incorrectly
### Acceptance criteria

The list of acceptance criteria (defined in the corresponding User Story or coming from your functional knowledge of the product)
These acceptance criteria must be met by the corrective implementation done by the developers.

### Context

iPhone XR

### Additional information

frequency, log files,...

PR #175, opened 2021-02-05
LA-174: Display upload screen correctly when upload from outside

manhduc-nguyen:
Refer to this bug: [https://github.com/linagora/linshare-mobile-flutter-app/issues/174](https://github.com/linagora/linshare-mobile-flutter-app/issues/174)

### DIFF
--- a/lib/presentation/widget/upload_file/upload_file_widget.dart
+++ b/lib/presentation/widget/upload_file/upload_file_widget.dart
@@ -79,7 +79,9 @@ class _UploadFileWidgetState extends State<UploadFileWidget> {
     uploadFileViewModel.setUploadFilesArgument(arguments.uploadFiles);
     uploadFileViewModel.setShareTypeArgument(arguments.shareType);
     uploadFileViewModel.setDocumentsArgument(arguments.documents);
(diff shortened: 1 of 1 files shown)


lib/presentation/widget/upload_file/upload_file_widget.dart
```diff
@@ -79,7 +79,9 @@ class _UploadFileWidgetState extends State<UploadFileWidget> {
     uploadFileViewModel.setUploadFilesArgument(arguments.uploadFiles);
     uploadFileViewModel.setShareTypeArgument(arguments.shareType);
     uploadFileViewModel.setDocumentsArgument(arguments.documents);
-    uploadFileViewModel.setWorkGroupDocumentUploadInfoArgument(arguments.workGroupDocumentUploadInfo);
+    if (arguments.workGroupDocumentUploadInfo != null) {
+      uploadFileViewModel.setWorkGroupDocumentUploadInfoArgument(arguments.workGroupDocumentUploadInfo);
+    }
```
hoangdat, line 84: can you explain the root cause of this?

manhduc-nguyen: Oh the `build(BuildContext context)` method is called again, and set the argument to `null` while we already had data inside `workGroupDocumentUploadInfoArgument `

hoangdat (approved):

PR #196, opened 2021-02-22
LA-184 Copy from Received Shares to My Space

imGok:
copyfromreceivedsharessuccess


domain/lib/src/usecases/share/share_document_view_state.dart
```diff
@@ -35,8 +35,6 @@ import 'package:domain/domain.dart';
 class ShareDocumentViewState extends ViewState {
   final List<Share> sharedRecipients;
 
-  Document get sharedDocument => sharedRecipients.first?.document;
```
manhduc-nguyen, line 38: `Share` class is using also when you share a file. So document is returned in this case. But actually we don't use it anywhere for now. So should we delete it? @hoangdat

lib/presentation/util/toast_message_handler.dart
```diff
@@ -168,10 +171,10 @@ class ToastMessageHandler {
         appToast.showErrorToast(AppLocalizations.of(context).files_could_not_be_deleted);
         _cleanSharedSpaceViewState();
       } else if (failure is CopyToMySpaceFailure) {
-        appToast.showToast(AppLocalizations.of(context).the_file_could_not_be_copied);
+        appToast.showErrorToast(AppLocalizations.of(context).the_file_could_not_be_copied);
```
manhduc-nguyen, line 174: ahhh thanks for this 💯

manhduc-nguyen:
LGTM! just confusing about `document` in `Share` class

domain/lib/src/model/share/share.dart
```diff
@@ -45,21 +44,38 @@ class Share extends Equatable {
   final String name;
   final DateTime creationDate;
   final DateTime modificationDate;
-  final Document document;
```
hoangdat, line 48: why we delete it?

hoangdat: if you do that, we will fail in sharing file

hoangdat: IMO, the better way is creating a separate model class for Received Share item

domain/lib/src/usecases/share/share_document_view_state.dart
hoangdat: Good idea. please discuss it in daily meeting

imGok:
@hoangdat @manhduc-nguyen I've created 2 models for Received Shares and Shares, review again please :)

domain/test/usecases/received/copy_multiple_files_to_my_space_from_received_shares_interactor_test.dart
```diff
@@ -0,0 +1,112 @@
...
+      .thenAnswer((_) async => [document2]);
+
+      final result = await _copyMultipleFilesFromReceivedSharesToMySpaceInteractor.execute(shares: [receivedShare1, receivedShare2]);
+      final resultsList = result
+          .map((success) => (success as CopyMultipleToMySpaceFromReceivedSharesAllSuccessViewState).resultList)
+          .getOrElse(() => []);
+      expect(resultsList, [Right<Failure, Success>(CopyToMySpaceViewState([document1])), Right<Failure, Success>(CopyToMySpaceViewState([document2]))]);
+    });
+
+    test('copy multiples files to my space from received shares interactor should return success with some failures', () async {
```
hoangdat, line 69: This case `copy multiples files to my space from received shares interactor should return success with some failures` and case `copy multiples files to my space from received shares interactor should fail with only failures` are quite versus.

hoangdat (approved):

hoangdat:
@manhduc-nguyen please take a look on this

manhduc-nguyen (approved):
LGTM

PR #198, opened 2021-02-23
LA-184 Multiple received shares copy

imGok:
multiplecopyreceiv


hoangdat:
LGTM too,

manhduc-nguyen (approved):
LGTM, have you also tested the share function ?

hoangdat (approved):

PR #202, opened 2021-02-25
LA-190 : Search in shared space

imGok:
screen_success

### DIFF
--- a/domain/lib/src/usecases/search_workgroup_node/search_workgroup_node_interactor.dart
+++ b/domain/lib/src/usecases/search_workgroup_node/search_workgroup_node_interactor.dart
@@ -1,0 +1,51 @@
+// LinShare is an open source filesharing software, part of the LinPKI software
+// suite, developed by Linagora.
+//
+// Copyright (C) 2021 LINAGORA
+//
+// This program is free software: you can redistribute it and/or modify it under the
+// terms of the GNU Affero General Public License as published by the Free Software
+// Foundation, either version 3 of the License, or (at your option) any later version,
+// provided you comply with the Additional Terms applicable for LinShare software by
+// Linagora pursuant to Section 7 of the GNU Affero General Public License,
+// subsections (b), (c), and (e), pursuant to which you must notably (i) retain the
+// display in the interface of the “LinShare™” trademark/logo, the "Libre & Free" mention,
+// the words “You are using the Free and Open Source version of LinShare™, powered by
+// Linagora © 2009–2020. Contribute to Linshare R&D by subscribing to an Enterprise
+// offer!”. You must also retain the latter notice in all asynchronous messages such as
+// e-mails sent with the Program, (ii) retain all hypertext links between LinShare and
+// http://www.linshare.org, between linagora.com and Linagora, and (iii) refrain from
+// infringing Linagora intellectual property rights over its trademarks and commercial
+// brands. Other Additional Terms apply, see
+// <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf>
+// for more details.
+// This program is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for
+// more details.
+// You should have received a copy of the GNU Affero General Public License and its
+// applicable Additional Terms for LinShare along with this program. If not, see
+// <http://www.gnu.org/licenses/> for the GNU Affero General Public License version
+//  3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf> for
+//  the Additional Terms applicable to LinShare software.
+//
+
+import 'package:dartz/dartz.dart';
+import 'package:domain/domain.dart';
+import 'package:domain/src/usecases/search_workgroup_node/search_workgroup_node_view_state.dart';
+
+class SearchWorkGroupNodeInteractor {
+
+  Future<Either<Failure, Success>> execute(List<WorkGroupNode> workGroupNodesList, SearchQuery searchQuery) async {
+    try {
+      final resultList = workGroupNodesList
+          .where((element) => element.name
+              .toLowerCase()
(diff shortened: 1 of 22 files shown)


lib/presentation/widget/home/home_app_bar.dart
```diff
@@ -180,6 +180,17 @@ class HomeAppBar extends StatelessWidget implements PreferredSizeWidget {
     }
   }
 
+  String _getSearchInt(BuildContext context, SearchDestination searchDestination) {
```
manhduc-nguyen, line 183: should be `_getSearchHintText`

lib/presentation/widget/home/home_viewmodel.dart
```diff
@@ -117,17 +118,27 @@ class HomeViewModel extends BaseViewModel {
 
   void cancelSearchState() {
     store.dispatch(DisableSearchStateAction());
-    if (store.state.uiState.searchState.searchDestination == SearchDestination.mySpace) {
+    if (store.state.uiState.searchState.searchDestination == SearchDestination.MY_SPACE) {
       store.dispatch(MySpaceAction(Right(DisableSearchViewState())));
       store.dispatch(CleanMySpaceStateAction());
     }
+
+    if (store.state.uiState.searchState.searchDestination == SearchDestination.SHARED_SPACE) {
```
manhduc-nguyen, line 126: pls use `else if (store.state.uiState.searchState.searchDestination == SearchDestination.SHARED_SPACE)` to optimize code

lib/presentation/widget/home/home_viewmodel.dart
```diff
@@ -117,17 +118,27 @@ class HomeViewModel extends BaseViewModel {
...
   }
 
   void search(String text) {
-    if (store.state.uiState.searchState.searchDestination == SearchDestination.mySpace) {
+    if (store.state.uiState.searchState.searchDestination == SearchDestination.MY_SPACE) {
       store.dispatch(MySpaceAction(Right(SearchDocumentNewQuery(SearchQuery(text.trim())))));
       store.dispatch(CleanMySpaceStateAction());
     }
+
+    if (store.state.uiState.searchState.searchDestination == SearchDestination.SHARED_SPACE) {
```
manhduc-nguyen, line 138: same, pls use `else if`

manhduc-nguyen:
@imGok you should also hide the RefreshIndicator while we in search state.

imGok:
> @imGok you should also hide the RefreshIndicator while we in search state.

Why hide refresh while searching ?

lib/presentation/widget/home/home_app_bar.dart
imGok: Oupsy!

lib/presentation/widget/home/home_viewmodel.dart
imGok: You're right thx

domain/lib/src/usecases/search_workgroup_node/search_workgroup_node_interactor.dart
```diff
@@ -0,0 +1,51 @@
...
+class SearchWorkGroupNodeInteractor {
+
+  Future<Either<Failure, Success>> execute(List<WorkGroupNode> workGroupNodesList, SearchQuery searchQuery) async {
+    try {
+      final resultList = workGroupNodesList
+          .where((element) => element.name
+              .toLowerCase()
+              .contains(searchQuery.value.toLowerCase()))
+          .toList();
+      return Right<Failure, Success>(SearchWorkGroupNodeSuccess(resultList));
```
hoangdat, line 46: no state for the case: `No Result`?
Btw, we need it for the screen
image
Am I right?

manhduc-nguyen:
> > @imGok you should also hide the RefreshIndicator while we in search state.
> 
> Why hide refresh while searching ?

Cause I think that there is no use of that in search feature. User can refresh for nothing, right. It's just the same result :D

domain/lib/src/usecases/search_workgroup_node/search_workgroup_node_interactor.dart
manhduc-nguyen: yeah, he is missing `No results found` text.

manhduc-nguyen: and I think we don't need state for `No Result`. `SearchWorkGroupNodeSuccess` with an empty result list is enough IMO.

lib/presentation/redux/states/ui_state.dart
```diff
@@ -73,26 +73,27 @@ class SearchState {
...
 
 enum SearchStatus {
-  active, inactive
+  ACTIVE, INACTIVE
 }
 
 enum SearchDestination {
-  mySpace
+  MY_SPACE,
+  SHARED_SPACE
```
manhduc-nguyen, line 98: I think this should be smthg like `SHARED_SPACE_NODE`, because we also have search feature from root folder, which actually is shared space u know.

lib/presentation/redux/actions/shared_space_action.dart
```diff
@@ -66,3 +66,10 @@ class EnableUploadButtonAction extends ActionOffline {
 class DisableUploadButtonAction extends ActionOffline {
   DisableUploadButtonAction();
 }
+
+@immutable
+class SharedSpaceNodesSetSearchResultAction extends ActionOffline {
```
manhduc-nguyen, line 71: where do u use this action ?

domain/lib/src/usecases/search_workgroup_node/search_workgroup_node_interactor.dart
hoangdat: okie for me

lib/presentation/redux/actions/shared_space_action.dart
imGok: Oupsy!! you are right

imGok: deleted it

manhduc-nguyen (approved):

hoangdat (approved):

PR #208, opened 2021-02-26
LA-185: Export files in work group

manhduc-nguyen:
export_shared_space

### DIFF
--- a/data/lib/src/datasource_impl/shared_space_document_datasource_impl.dart
+++ b/data/lib/src/datasource_impl/shared_space_document_datasource_impl.dart
@@ -30,6 +30,7 @@
 //  the Additional Terms applicable to LinShare software.
 //
 
+import 'dart:async';
 import 'dart:io';
 
 import 'package:data/data.dart';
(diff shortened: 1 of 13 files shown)


data/lib/src/datasource_impl/shared_space_document_datasource_impl.dart
```diff
@@ -161,4 +162,68 @@ class SharedSpaceDocumentDataSourceImpl implements SharedSpaceDocumentDataSource
...
+                .withPathParameter(workgroupNode.sharedSpaceId.uuid)
+                .withPathParameter('nodes')
+                .withPathParameter(workgroupNode.workGroupNodeId.uuid)
+                .withPathParameter(Endpoint.download)
+                .generateEndpointPath(),
+            cancelToken,
+            token),
+        getTemporaryDirectory()
+      ]).then((values) {
+        final fileStream = (values[0] as ResponseBody).stream;
```
imGok, line 182: WDYT about testing if values is not empty ?

domain/lib/src/usecases/shared_space/download_multiple_node_ios_interactor.dart
```diff
@@ -0,0 +1,65 @@
...
+        _downloadNodeIOSInteractor.execute(element, cancelToken)));
+    if (listResult.length == 1) {
+      return listResult.first;
+    } else {
+      var failedFileCount = 0;
+      listResult.forEach((element) {
+        if (element is Left) {
+          failedFileCount++;
+        }
+      });
```
imGok, line 56: `.whereType<Left>()` is so clean you should use it !

manhduc-nguyen: nice

hoangdat (approved):

data/lib/src/datasource_impl/shared_space_document_datasource_impl.dart
manhduc-nguyen: Oh I don't think we need that because we already have `try catch` for that.

domain/lib/src/usecases/shared_space/download_multiple_node_ios_interactor.dart
manhduc-nguyen: @hoangdat @imGok fixed and rebased, pls check again

imGok (approved):

PR #213, opened 2021-03-01
LA-192 Delete Shared Space

imGok:
deletework

### DIFF
--- a/domain/lib/src/model/sharedspace/shared_space_role.dart
+++ b/domain/lib/src/model/sharedspace/shared_space_role.dart
@@ -40,6 +40,10 @@ class SharedSpaceRole extends Equatable {
 
   SharedSpaceRole(this.sharedSpaceRoleId, this.name);
 
+  factory SharedSpaceRole.initial() {
+    return SharedSpaceRole(SharedSpaceRoleId(''), SharedSpaceRoleName.READER);
+  }
+
   @override
   List<Object> get props => [sharedSpaceRoleId, name];
 }
--- a/domain/lib/src/usecases/shared_space/remove_shared_space_interactor.dart
+++ b/domain/lib/src/usecases/shared_space/remove_shared_space_interactor.dart
@@ -28,11 +28,28 @@
 // <http://www.gnu.org/licenses/> for the GNU Affero General Public License version
 //  3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf> for
 //  the Additional Terms applicable to LinShare software.
+//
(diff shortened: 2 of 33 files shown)


domain/lib/src/usecases/shared_space/remove_shared_space_interactor.dart
```diff
@@ -0,0 +1,51 @@
...
+
+  RemoveSharedSpaceInteractor(this._sharedSpaceRepository);
+
+  Future<Either<Failure, Success>> execute(
+    SharedSpaceId sharedSpaceId,
+  ) async {
+    try {
+      final sharedSpace = await _sharedSpaceRepository.deleteSharedSpace(sharedSpaceId);
+      return Right<Failure, Success>(DeleteSharedSpaceViewState(sharedSpace));
+    } catch (exception) {
```
hoangdat, line 47: how about the case `404` - shared space not found?

imGok: WDYM ? It's thrown by datasource and catch into this

hoangdat: I want for this case, We don't need to show toast for user, only reload the list

imGok: You mean if 404 is thrown => no toast ?

imGok:
@manhduc-nguyen @hoangdat finished you can review

manhduc-nguyen (approved):

domain/lib/src/model/sharedspace/shared_space_role.dart
```diff
@@ -40,6 +40,10 @@ class SharedSpaceRole extends Equatable {
 
   SharedSpaceRole(this.sharedSpaceRoleId, this.name);
 
+  factory SharedSpaceRole.initial() {
+    return SharedSpaceRole(SharedSpaceRoleId('unknown'), SharedSpaceRoleName.UNKNOWN);
```
hoangdat, line 44: why not use `Reader` instead?

domain/lib/src/usecases/shared_space/remove_shared_space_interactor.dart
hoangdat: yes.

imGok: Done

hoangdat (approved):

PR #220, opened 2021-03-03
LA-186 Download WorkGroupNode in presentation

hoangdat:
- presentation layer for download nodes

### DIFF
--- a/lib/presentation/widget/shared_space/file_surfing/workgroup_nodes_surfing_viewmodel.dart
+++ b/lib/presentation/widget/shared_space/file_surfing/workgroup_nodes_surfing_viewmodel.dart
@@ -56,6 +56,7 @@ import 'package:linshare_flutter_app/presentation/view/modal_sheets/confirm_moda
 import 'package:linshare_flutter_app/presentation/widget/shared_space/file_surfing/workgroup_nodes_surfing_state.dart';
 import 'package:linshare_flutter_app/presentation/widget/base/base_viewmodel.dart';
 import 'package:linshare_flutter_app/presentation/widget/shared_space/file_surfing/workgroup_nodes_surfling_arguments.dart';
+import 'package:permission_handler/permission_handler.dart';
 import 'package:redux/src/store.dart';
 import 'package:redux_thunk/redux_thunk.dart';
 import 'package:rxdart/rxdart.dart';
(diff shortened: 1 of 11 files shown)


lib/presentation/widget/shared_space/file_surfing/workgroup_nodes_surfing_viewmodel.dart
```diff
@@ -356,6 +359,41 @@ class WorkGroupNodesSurfingViewModel extends BaseViewModel {
     };
   }
 
+  OnlineThunkAction downloadNodeAction(List<WorkGroupNode> nodes, {ItemSelectionType itemSelectionType = ItemSelectionType.single}) {
```
manhduc-nguyen, line 362: if you return an `OnlineThunkAction` then you have to call `store.dispatch(yourAction)` to execute it. But in this case this should be a normal `void` function `void downloadNodeAction(...)` to be called from widget

lib/presentation/widget/shared_space/file_surfing/workgroup_nodes_surfing_viewmodel.dart
```diff
@@ -356,6 +359,41 @@ class WorkGroupNodesSurfingViewModel extends BaseViewModel {
...
+        }
+      }
+    });
+  }
+
+  void _downloadNodes(List<WorkGroupNode> nodes, {ItemSelectionType itemSelectionType = ItemSelectionType.single}) {
+    store.dispatch(_handleDownloadNodes(nodes.map((node) => node.workGroupNodeId)));
+  }
+
+  ThunkAction<AppState> _handleDownloadNodes(List<WorkGroupNodeId> nodeIds) {
```
manhduc-nguyen, line 388: this should be `OnlineThunkAction`

lib/presentation/widget/shared_space/file_surfing/workgroup_nodes_surfing_viewmodel.dart
```diff
@@ -356,6 +359,41 @@ class WorkGroupNodesSurfingViewModel extends BaseViewModel {
...
+            default: _appNavigation.popBack();
+            break;
+          }
+        }
+      }
+    });
+  }
+
+  void _downloadNodes(List<WorkGroupNode> nodes, {ItemSelectionType itemSelectionType = ItemSelectionType.single}) {
+    store.dispatch(_handleDownloadNodes(nodes.map((node) => node.workGroupNodeId)));
```
manhduc-nguyen, line 385: I think you are missing this. We have to cancel multi selection on UI after they have pressed download button
`if (itemSelectionType == ItemSelectionType.multiple) {
      cancelSelection();
    }
`

manhduc-nguyen (approved):
LGTM

Issue #224, opened 2021-03-04
Lack of actions in context menu of multiple selected files


hoangdat:
### Description

Lack of actions in context menu of multiple selected files

### Expected result

![Screen Shot 2021-03-04 at 11.30.49.png](https://images.zenhubusercontent.com/5f7d8ef759d36aecc684e2a4/8eed1902-e1ce-4ab3-9980-192fb73eb7b3)

### Current behavior

All actions in first screen not show in next screen

#### Reproduction Steps

Select files in MySpace -> Click on three dot of context menu -> check the list of actions

PR #226, opened 2021-03-04
LA-224: add missing actions to bottom bar and more actions popup

manhduc-nguyen:
refer to this bug [https://github.com/linagora/linshare-mobile-flutter-app/issues/224](https://github.com/linagora/linshare-mobile-flutter-app/issues/224)

Screen Shot 2021-03-04 at 17 33 37Screen Shot 2021-03-04 at 17 33 43

### DIFF
--- a/lib/presentation/widget/myspace/my_space_viewmodel.dart
+++ b/lib/presentation/widget/myspace/my_space_viewmodel.dart
@@ -188,10 +188,10 @@ class MySpaceViewModel extends BaseViewModel {
   }
 
   void shareDocuments(List<Document> documents, {ItemSelectionType itemSelectionType = ItemSelectionType.single}) {
-    if (itemSelectionType == ItemSelectionType.single) {
-      _appNavigation.popBack();
+    _appNavigation.popBack();
+    if (itemSelectionType == ItemSelectionType.multiple) {
+      cancelSelection();
     }
-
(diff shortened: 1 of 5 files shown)


hoangdat:
Please do it for SharedSpace and SSD and RS also

lib/presentation/widget/myspace/my_space_viewmodel.dart
```diff
@@ -188,10 +192,12 @@ class MySpaceViewModel extends BaseViewModel {
   }
 
   void shareDocuments(List<Document> documents, {ItemSelectionType itemSelectionType = ItemSelectionType.single}) {
-    if (itemSelectionType == ItemSelectionType.single) {
+    if (_appNavigation.canPop()) {
```
hoangdat, line 195: what is the purpose of it?

hoangdat (approved):

lib/presentation/widget/myspace/my_space_viewmodel.dart
manhduc-nguyen: I want to make sure that we only pop out the widget if its possible, and I think moving this directly inside the `void popBack()` function is better, shorter code to write. let me change this :)

manhduc-nguyen:
@imGok @hoangdat done, also add missing actions for `SharedSpace too`

imGok (approved):

PR #261, opened 2021-03-17
LA-229: Preview file in MySpace

manhduc-nguyen:
Pls review this, I will add unit test later.

demo_preview


domain/lib/src/model/preview/document_uti.dart
```diff
@@ -0,0 +1,36 @@
...
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for
+// more details.
+// You should have received a copy of the GNU Affero General Public License and its
+// applicable Additional Terms for LinShare along with this program. If not, see
+// <http://www.gnu.org/licenses/> for the GNU Affero General Public License version
+//  3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf> for
+//  the Additional Terms applicable to LinShare software.
+
+class DocumentUti {
```
imGok, line 32: What does `uti` stand for please ?

manhduc-nguyen: its [Uniform Type Identifier](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/understanding_utis/understand_utis_conc/understand_utis_conc.html) in iOS, for short explanation, its just like `mime type` but for Apple.

imGok: Ok thx

imGok (approved):

dab246:
Good job :)

huynguyennovem:
look good from me

data/lib/src/network/linshare_download_manager.dart
```diff
@@ -0,0 +1,105 @@
...
+
+    try {
+      await Future.wait([
+        _linShareHttpClient.downloadFile(downloadUrl, cancelToken, permanentToken),
+        directoryToSave]).then((values) {
+        final fileStream = (values[0] as ResponseBody).stream;
+        final tempFilePath = '${(values[1] as Directory).absolute.path}/${filename}';
+
+        final file = File(tempFilePath);
+        file.createSync(recursive: true);
```
hoangdat, line 63: what is this? Why we need this?

manhduc-nguyen:
```
* If [recursive] is false, the default, the file is created
   * only if all directories in the path exist.
   * If [recursive] is true, all non-existing path components are created.

```

Basically, it will create directories from the file path if any of it does not exist

huynguyennovem (approved):
/approve

hoangdat (approved):

PR #278, opened 2021-03-23
LA-228: Share with device contact

manhduc-nguyen:
https://user-images.githubusercontent.com/16035728/112450939-7ea86a80-8d87-11eb-983f-3fb65be42289.mp4

### DIFF
--- a/data/lib/src/datasource_impl/contact_datasource_impl.dart
+++ b/data/lib/src/datasource_impl/contact_datasource_impl.dart
@@ -1,0 +1,52 @@
+// LinShare is an open source filesharing software, part of the LinPKI software
+// suite, developed by Linagora.
+//
+// Copyright (C) 2021 LINAGORA
+//
+// This program is free software: you can redistribute it and/or modify it under the
+// terms of the GNU Affero General Public License as published by the Free Software
+// Foundation, either version 3 of the License, or (at your option) any later version,
+// provided you comply with the Additional Terms applicable for LinShare software by
+// Linagora pursuant to Section 7 of the GNU Affero General Public License,
+// subsections (b), (c), and (e), pursuant to which you must notably (i) retain the
+// display in the interface of the “LinShare™” trademark/logo, the "Libre & Free" mention,
+// the words “You are using the Free and Open Source version of LinShare™, powered by
+// Linagora © 2009–2021. Contribute to Linshare R&D by subscribing to an Enterprise
+// offer!”. You must also retain the latter notice in all asynchronous messages such as
+// e-mails sent with the Program, (ii) retain all hypertext links between LinShare and
+// http://www.linshare.org, between linagora.com and Linagora, and (iii) refrain from
+// infringing Linagora intellectual property rights over its trademarks and commercial
+// brands. Other Additional Terms apply, see
+// <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf>
+// for more details.
+// This program is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for
+// more details.
+// You should have received a copy of the GNU Affero General Public License and its
+// applicable Additional Terms for LinShare along with this program. If not, see
+// <http://www.gnu.org/licenses/> for the GNU Affero General Public License version
+//  3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf> for
+//  the Additional Terms applicable to LinShare software.
+
+import 'package:contacts_service/contacts_service.dart' as contact_service;
+import 'package:data/src/datasource/contact_datasource.dart';
+import 'package:domain/domain.dart';
+import 'package:domain/src/model/autocomplete/autocomplete_pattern.dart';
+import 'package:domain/src/model/contact/contact.dart' as contact;
+
+class ContactDataSourceImpl extends ContactDataSource {
+
+  @override
+  Future<List<contact.Contact>> getContactSuggestions(AutoCompletePattern autoCompletePattern) async {
+    final suggestedList = await contact_service.ContactsService.getContactsByEmailOrName(autoCompletePattern.value);
+    return suggestedList.expand((contact) => _toDeviceContact(contact)).toList();
+  }
+
+  List<DeviceContact> _toDeviceContact(contact_service.Contact contact) {
+    if (contact.emails != null) {
+      return contact.emails.map((email) => DeviceContact(contact.displayName, email.value)).toList();
+    }
+    return <DeviceContact>[];
+  }
+}
--- a/domain/lib/src/model/contact/contact.dart
+++ b/domain/lib/src/model/contact/contact.dart
@@ -1,7 +1,7 @@
 // LinShare is an open source filesharing software, part of the LinPKI software
 // suite, developed by Linagora.
 //
-// Copyright (C) 2020 LINAGORA
+// Copyright (C) 2021 LINAGORA
 //
 // This program is free software: you can redistribute it and/or modify it under the
 // terms of the GNU Affero General Public License as published by the Free Software
@@ -11,7 +11,7 @@
 // subsections (b), (c), and (e), pursuant to which you must notably (i) retain the
 // display in the interface of the “LinShare™” trademark/logo, the "Libre & Free" mention,
 // the words “You are using the Free and Open Source version of LinShare™, powered by
-// Linagora © 2009–2020. Contribute to Linshare R&D by subscribing to an Enterprise
+// Linagora © 2009–2021. Contribute to Linshare R&D by subscribing to an Enterprise
 // offer!”. You must also retain the latter notice in all asynchronous messages such as
 // e-mails sent with the Program, (ii) retain all hypertext links between LinShare and
 // http://www.linshare.org, between linagora.com and Linagora, and (iii) refrain from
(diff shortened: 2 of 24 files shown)


domain/lib/src/model/contact/contact.dart
```diff
@@ -0,0 +1,42 @@
...
+// more details.
+// You should have received a copy of the GNU Affero General Public License and its
+// applicable Additional Terms for LinShare along with this program. If not, see
+// <http://www.gnu.org/licenses/> for the GNU Affero General Public License version
+//  3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf> for
+//  the Additional Terms applicable to LinShare software.
+
+import 'package:equatable/equatable.dart';
+
+abstract class Contact with EquatableMixin {
```
hoangdat, line 34: as I read before, we can use Mixin to replace interface, please investigate on that.

imGok (approved):

manhduc-nguyen:
@imGok @hoangdat pls check again, also added presentation layer

domain/lib/src/model/contact/contact.dart
manhduc-nguyen: yes, I took a research. We can easily use `mixin` to replace `abstract`. as I read: 

> In object-oriented programming languages, a Mixin is a class that contains methods for use by other classes without having to be the parent class of those other classes 

Also there is other benefit of `mixin` is to avoid `Deadly Diamond of Dead (DDD)`, you can read more at this [medium article](https://medium.com/flutter-community/https-medium-com-shubhamhackzz-dart-for-flutter-mixins-in-dart-f8bb10a3d341).  So basically, I see that there is no different to using either of them in our project. I think we don't need `mixin` for now.

lib/presentation/widget/upload_file/upload_file_viewmodel.dart
```diff
@@ -48,23 +48,31 @@ import 'package:linshare_flutter_app/presentation/widget/destination_picker/dest
...
 import 'package:linshare_flutter_app/presentation/util/extensions/media_type_extension.dart';
+import 'package:permission_handler/permission_handler.dart';
 import 'package:redux/redux.dart';
 import 'package:redux_thunk/redux_thunk.dart';
 import 'package:rxdart/rxdart.dart';
 
 import 'upload_destination_type.dart';
 
+enum ContactSuggestionSource {
+  all, linShareContact, deviceContact
```
imGok, line 59: Can you use enum syntaxe like "ALL, LINSHARE_CONTACT ..." pls ?

lib/presentation/widget/upload_file/upload_file_viewmodel.dart
```diff
@@ -86,6 +94,8 @@ class UploadFileViewModel extends BaseViewModel {
           break;
       }
     });
+
+    Future.delayed(Duration(milliseconds: 500), () => _checkContactPermission());
```
imGok, line 98: What is it for ? You want to check after widget building?

lib/presentation/widget/upload_file/upload_file_viewmodel.dart
manhduc-nguyen: oops, bad habit for me

data/lib/src/datasource_impl/contact_datasource_impl.dart
```diff
@@ -0,0 +1,57 @@
...
+import 'package:domain/src/model/autocomplete/autocomplete_pattern.dart';
+import 'package:domain/src/model/contact/contact.dart' as contact;
+
+class ContactDataSourceImpl extends ContactDataSource {
+
+  @override
+  Future<List<contact.Contact>> getContactSuggestions(AutoCompletePattern autoCompletePattern) async {
+    final suggestedList = await ContactsService.getContactsByEmailOrName(autoCompletePattern.value);
+    final deviceContactList = <contact.Contact>[];
+    suggestedList.forEach((suggestion) {
```
dab246, line 44: WDYT, if using map.

lib/presentation/widget/upload_file/upload_file_viewmodel.dart
manhduc-nguyen: yes, basically we have to show the request permission dialog after the widget has finished building layout. tried all solutions in here [https://stackoverflow.com/questions/49466556/flutter-run-method-on-widget-build-complete](https://stackoverflow.com/questions/49466556/flutter-run-method-on-widget-build-complete) but it will end up like this.

Screen Shot 2021-03-25 at 11 45 09

hoangdat: cool.

domain/lib/src/model/contact/contact.dart
hoangdat: I agree for the case of `Contact` and POJO class

data/lib/src/datasource_impl/contact_datasource_impl.dart
manhduc-nguyen: oh you mean using map here to return a different type directly? I don't think that will work in my case. I also need to check if the contact return from the library have more than 1 email, then I have to split that into 2 different `DeviceContact`

domain/lib/src/usecases/autocomplete/get_autocomplete_sharing_with_device_contact_interactor.dart
```diff
@@ -0,0 +1,66 @@
...
+      final resultExecutions = await Future.wait([
+        _getAutoCompleteSharingInteractor.execute(autoCompletePattern, autoCompleteType),
+        _getDeviceContactSuggestionsInteractor.execute(autoCompletePattern)]);
+
+      final autoCompleteResults = resultExecutions.first.fold(
+        (failure) => <AutoCompleteResult>[],
+        (success) => success is AutoCompleteViewState ? success.results : <AutoCompleteResult>[]);
+
+      resultExecutions.last.fold(
+        (failure) => null,
```
hoangdat, line 54: what is null, how about other `Failure`

domain/lib/src/model/contact/contact.dart
manhduc-nguyen: yes, especially for POJO class because `mixin` should not be considered as a return type.

data/lib/src/datasource_impl/contact_datasource_impl.dart
hoangdat: `flatMap` then convert to deviceContact. I don't like to add each item to a shared list.

lib/presentation/widget/upload_file/upload_file_viewmodel.dart
```diff
@@ -289,6 +308,18 @@ class UploadFileViewModel extends BaseViewModel {
...
+    final permissionStatus = await Permission.contacts.status;
+    if (permissionStatus.isGranted) {
+      _contactSuggestionSource = ContactSuggestionSource.all;
+    } else if (!permissionStatus.isPermanentlyDenied) {
+      final requestedPermission = await Permission.contacts.request();
+      _contactSuggestionSource = requestedPermission == PermissionStatus.granted
+          ? ContactSuggestionSource.all
+          : _contactSuggestionSource;
+    }
+  }
```
hoangdat, line 321: I did not see the case for `permanentDenied`, so that, user can not use this if he makes any mistake decision

data/lib/src/datasource_impl/contact_datasource_impl.dart
manhduc-nguyen: thank guys, I think its better now.

lib/presentation/widget/upload_file/upload_file_viewmodel.dart
manhduc-nguyen: done

domain/lib/src/usecases/autocomplete/get_autocomplete_sharing_with_device_contact_interactor.dart
manhduc-nguyen: in case of getting suggestion from device contact failed, we won't add anything to `autoCompleteResults` list and still returns result from remote contact suggestions. Thats my view intend here.

manhduc-nguyen:
@imGok @hoangdat @dab246 pls review again guys

domain/lib/src/usecases/autocomplete/get_autocomplete_sharing_with_device_contact_interactor.dart
hoangdat: okie, how about return `EMPTY_CONTACT = emptyList()` instead

data/lib/src/datasource_impl/contact_datasource_impl.dart
```diff
@@ -40,18 +40,15 @@ class ContactDataSourceImpl extends ContactDataSource {
...
+    return suggestedList.expand((suggestion) {
       if (suggestion.emails != null && suggestion.emails.isNotEmpty) {
         if (suggestion.emails.length == 1) {
-          deviceContactList.add(DeviceContact(suggestion.displayName, suggestion.emails.first.value));
+          return [DeviceContact(suggestion.displayName, suggestion.emails.first.value)];
         } else {
-          suggestion.emails.forEach((item) {
-            deviceContactList.add(DeviceContact(suggestion.displayName, item.value));
-          });
+          return suggestion.emails.map((email) => DeviceContact(suggestion.displayName, email.value)).toList();
```
hoangdat: Please split it to another method.

lib/presentation/widget/upload_file/upload_file_viewmodel.dart
imGok: It depends to much on users device but if it's the only way to do it let's go

imGok (approved):

hoangdat (approved):

PR #288, opened 2021-03-25
LA-247 Work Group members

imGok:
image

### DIFF
--- a/data/lib/src/datasource_impl/shared_space_member_datasource_impl.dart
+++ b/data/lib/src/datasource_impl/shared_space_member_datasource_impl.dart
@@ -1,0 +1,60 @@
+// LinShare is an open source filesharing software, part of the LinPKI software
+// suite, developed by Linagora.
+//
+// Copyright (C) 2020 LINAGORA
+//
+// This program is free software: you can redistribute it and/or modify it under the
+// terms of the GNU Affero General Public License as published by the Free Software
+// Foundation, either version 3 of the License, or (at your option) any later version,
+// provided you comply with the Additional Terms applicable for LinShare software by
+// Linagora pursuant to Section 7 of the GNU Affero General Public License,
+// subsections (b), (c), and (e), pursuant to which you must notably (i) retain the
+// display in the interface of the “LinShare™” trademark/logo, the "Libre & Free" mention,
+// the words “You are using the Free and Open Source version of LinShare™, powered by
+// Linagora © 2009–2020. Contribute to Linshare R&D by subscribing to an Enterprise
+// offer!”. You must also retain the latter notice in all asynchronous messages such as
+// e-mails sent with the Program, (ii) retain all hypertext links between LinShare and
+// http://www.linshare.org, between linagora.com and Linagora, and (iii) refrain from
+// infringing Linagora intellectual property rights over its trademarks and commercial
+// brands. Other Additional Terms apply, see
+// <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf>
+// for more details.
+// This program is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for
+// more details.
+// You should have received a copy of the GNU Affero General Public License and its
+// applicable Additional Terms for LinShare along with this program. If not, see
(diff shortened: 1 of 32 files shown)


hoangdat:
go ahead, quite good to me

data/lib/src/datasource_impl/shared_space_member_datasource_impl.dart
```diff
@@ -0,0 +1,60 @@
...
+
+  @override
+  Future<List<SharedSpaceMember>> getMembers(SharedSpaceId sharedSpaceId) {
+    return Future.sync(() async {
+        return (await _linShareHttpClient.getSharedSpaceMembers(sharedSpaceId)).map((memberDto) => memberDto.toSharedSpaceMember()).toList();
+      }).catchError((error) {
+        _remoteExceptionThrower.throwRemoteException(error, handler: (DioError error) {
+          if (error.response.statusCode == 404) {
+            throw SharedSpacesNotFound();
+          } if (error.response.statusCode == 403) {
```
dab246, line 52: You should substitute else if for if

domain/test/usecases/shared_space/get_all_shared_space_members_interactor_test.dart
```diff
@@ -0,0 +1,76 @@
...
+      sharedSpaceMemberRepository = MockSharedSpaceMemberRepository();
+      _getAllSharedSpaceMembersInteractor = GetAllSharedSpaceMembersInteractor(sharedSpaceMemberRepository);
+    });
+
+    test('getAllSharedSpaceMembersInteractor should return success with members list', () async {
+      when(sharedSpaceMemberRepository.getMembers(sharedSpaceId1)).thenAnswer((_) async => [sharedSpaceMember1, sharedSpaceMember2]);
+
+      final result = await _getAllSharedSpaceMembersInteractor.execute(sharedSpaceId1);
+
+      final membersList = result.map((success) => (success as SharedSpaceMembersViewState).members)
```
dab246, line 56: You should use is change because as:Ω`success is SharedSpaceMembersViewState ? success.members`

imGok: I want to return a exact list and It'll return null if there is any problem

imGok: I think `map((success) => success is SharedSpaceMembersViewState ? success.members : [])` is the same as mine

imGok:
@hoangdat @dab246 @manhduc-nguyen @huynguyennovem Avengers Assemble !

domain/lib/src/model/sharedspace/shared_space_role_name.dart
```diff
@@ -33,3 +33,9 @@
 enum SharedSpaceRoleName {
   READER, CONTRIBUTOR, WRITER, ADMIN
 }
+
+extension SharedSpaceRoleNameExtension on SharedSpaceRoleName {
+  String getEnumName() {
+    return toString().split('.').last.toLowerCase();
```
hoangdat, line 39: It is for presentation? If it is, please move this to presentation and get the name from the `app_localization`, we need to translate it to depend on the language which user selected

lib/presentation/widget/shared_space_details/shared_space_details_viewmodel.dart
```diff
@@ -65,10 +67,19 @@ class SharedSpaceDetailsViewModel extends BaseViewModel {
         .accountQuota;
   }
 
+  Future<List<SharedSpaceMember>> getSharedSpaceMembers(SharedSpaceId sharedSpaceId) async {
+    return (await _getAllSharedSpaceMembersInteractor.execute(sharedSpaceId))
+        .map((result) => result as SharedSpaceMembersViewState)
```
hoangdat, line 72: Casting with no check, I concern it too much.

manhduc-nguyen (approved):

domain/test/usecases/shared_space/get_all_shared_space_members_interactor_test.dart
dab246: If success is not an instance of SharedSpaceMembersViewState it will fire an exception while runtime

hoangdat:
Merge commit -> test resolve conflict online, its not good. SHOULD NOT DO IT IN THE FUTURE

hoangdat (approved):

hoangdat (approved):

PR #290, opened 2021-03-29
LA-248 Add Shared Space Member

imGok:
addMembers


domain/lib/src/model/sharedspace/shared_space_role_name.dart
```diff
@@ -33,3 +33,9 @@
 enum SharedSpaceRoleName {
```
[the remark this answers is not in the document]
imGok: I did but the JSON is in Upper case so it crashed, I need to find a solution for that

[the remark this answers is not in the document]
hoangdat: in the case of parsing Data, no need to follow the rule `lowerCamelCase `

hoangdat:
Please replace the destination from `linagora:master` to `linagora:maintenance` @imGok

imGok:
@hoangdat @dab246 @huynguyennovem Review please

domain/lib/src/model/autocomplete/autocomplete_result_type.dart
```diff
@@ -31,7 +31,7 @@
 //
 
 enum AutoCompleteResultType {
-  simple, user, mailingList
+  simple, user, mailingList, threadmember
```
dab246, line 34: You should use lowerCamelCase `threadMember`

hoangdat: it is type from response, so it is no problem with that

lib/presentation/widget/shared_space_details/add_shared_space_member/add_shared_space_member_viewmodel.dart
```diff
@@ -0,0 +1,111 @@
...
+
+  void addSharedSpaceMember(
+      SharedSpaceId sharedSpaceId, SharedSpaceMemberAutoCompleteResult autoCompleteResult) {
+    store.dispatch(
+        _addSharedSpaceAction(sharedSpaceId, AccountId(autoCompleteResult.userUuid.uuid)));
+  }
+
+  OnlineThunkAction _addSharedSpaceAction(SharedSpaceId sharedSpaceId, AccountId accountId) {
+    return OnlineThunkAction((Store<AppState> store) async {
+      final roles = (await _getAllSharedSpaceRolesInteractor.execute())
```
hoangdat, line 99: I don't think it is a good idea, `role name` and `role id` are not changed too much, why we not fetch it only 1 time when open the add member screen instead of fetch it every time add member?

lib/presentation/widget/shared_space_details/add_shared_space_member/add_shared_space_member_widget.dart
```diff
@@ -0,0 +1,267 @@
...
+  final TextEditingController _typeAheadController = TextEditingController();
+  StreamSubscription _streamSubscription;
+
+  @override
+  void initState() {
+    _streamSubscription =
+        _model.store.onChange.map((event) => event.sharedSpaceState.viewState).listen((viewState) {
+      viewState.fold((_) => null, (success) {
+        if (success is AddSharedSpaceMemberViewState) {
+          setState(() {});
```
hoangdat, line 67: it is so weird to me. Why not react it with `StoreConnector`?

lib/presentation/widget/shared_space_details/add_shared_space_member/add_shared_space_member_widget.dart
```diff
@@ -0,0 +1,267 @@
...
+          style: TextStyle(
+            color: _model.selectedRole == role
+                ? AppColor.primaryColor
+                : AppColor.addSharedSpaceMemberRoleTileColor,
+            fontSize: 16,
+          )),
+      leading: _model.selectedRole == role
+          ? Icon(Icons.check, color: AppColor.primaryColor)
+          : SizedBox(),
+      onTap: () => setState(() => _model.selectRole(role)),
```
hoangdat, line 264: so weird to me too.

hoangdat: I DON'T LIKE `setState()`. Please minimize to use that. @manhduc-nguyen @imGok @huynguyennovem @dab246

hoangdat (changes requested):

hoangdat:
`domain` and `data` quite good. But `presentation` is so weird to the convention of team. 
When you break the convention, everything will be change, the architecture is not consistent, IT IS VERY DANGEROUS for us to maintain the project.

lib/presentation/widget/shared_space_details/shared_space_details_widget.dart
```diff
@@ -219,53 +219,31 @@ class _SharedSpaceDetailsWidgetState extends State<SharedSpaceDetailsWidget> {
...
-                members[index].account.mail,
-                style: TextStyle(
-                  fontSize: 14,
-                  fontWeight: FontWeight.normal,
-                  fontStyle: FontStyle.italic,
-                  color: AppColor.documentModifiedDateItemTextColor
+  Widget _membersTabWidget(SharedSpaceDetailsInfo sharedSpaceDetails) {
+    return RefreshIndicator(
+        onRefresh: () async {
+          setState(() {});
```
huynguyennovem, line 225: `setState(() {});` will make all 3 tabs be refreshed then call all corresponding APIs.

huynguyennovem (commented):
IMO: Try to use `setState()` as little as possible. Consider [this](https://medium.com/@dabralrajat4/replacing-your-setstate-calls-with-valuenotifier-in-flutter-d35aab1c64db)

lib/presentation/widget/shared_space_details/add_shared_space_member/add_shared_space_member_widget.dart
hoangdat: We can not avoid it? @imGok

lib/presentation/widget/shared_space_details/add_shared_space_member/add_shared_space_member_viewmodel.dart
```diff
@@ -0,0 +1,125 @@
...
+      SharedSpaceId sharedSpaceId, SharedSpaceMemberAutoCompleteResult autoCompleteResult) {
+    store.dispatch(
+        _addSharedSpaceAction(sharedSpaceId, AccountId(autoCompleteResult.userUuid.uuid)));
+  }
+
+  OnlineThunkAction _addSharedSpaceAction(SharedSpaceId sharedSpaceId, AccountId accountId) {
+    return OnlineThunkAction((Store<AppState> store) async {
+      final selectedRoleId = store.state.sharedSpaceState.rolesList
+          .firstWhere((element) => element.name == selectedRole)
+          .sharedSpaceRoleId;
```
hoangdat, line 106: how about the case we can not get `roleIds`, this `selectedRoleId` is null?

domain/lib/src/model/sharedspace/shared_space_role.dart
```diff
@@ -38,12 +38,14 @@ class SharedSpaceRole extends Equatable {
   final SharedSpaceRoleId sharedSpaceRoleId;
   final SharedSpaceRoleName name;
 
-  SharedSpaceRole(this.sharedSpaceRoleId, this.name);
+  final bool enabled;
```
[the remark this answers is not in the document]
imGok: It's sent by API, I guess it tells if role is enabled or not

imGok:
We can not avoid it? @imGok

No we can't because the widget wont update otherwise, an other solution would be to store the current role value in a state but I don't want it's pointless

hoangdat:
> We can not avoid it? @imGok
> 
> No we can't because the widget wont update otherwise, an other solution would be to store the current role value in a state but I don't want it's pointless

I really want state for `add_member` also. Idealy, we should use `state` as much as possible. Only one thing to care in the app, it is `appState`. It is easier for us when debugging instead of finding somewhere to store something. One and only one: `AppState`. It is the point why we use Redux. @imGok @huynguyennovem @dab246 @manhduc-nguyen

imGok:
> I really want state for `add_member` also. Idealy, we should use `state` as much as possible. Only one thing to care in the app, it is `appState`. It is easier for us when debugging instead of finding somewhere to store something. One and only one: `AppState`. It is the point why we use Redux. @imGok @huynguyennovem @dab246 @manhduc-nguyen

For me, Redux is used to store persistent data across the app not view data (volatile data).
I don't want to use state for everything since onChanges listen to all state, so if we charge state more it will slow the app.
In our case, selectedRole is used only in this view.

I can try to find a solution with a stream builder tho

lib/presentation/widget/shared_space_details/add_shared_space_member/add_shared_space_member_viewmodel.dart
imGok: WDYT?
```dart
  OnlineThunkAction _addSharedSpaceAction(SharedSpaceId sharedSpaceId, AccountId accountId) {
    return OnlineThunkAction((Store<AppState> store) async {
      final rolesList = store.state.sharedSpaceState.rolesList;
      if (rolesList.isEmpty) {
        return;
      }

      final role = rolesList.firstWhere((element) => element.name == selectedRole, orElse: () => null);
      if (role == null) {
        return;
      }

      await _addSharedSpaceMemberInteractor
          .execute(
              sharedSpaceId, AddSharedSpaceMemberRequest(accountId, sharedSpaceId, role.sharedSpaceRoleId))
          .then((result) => store.dispatch(SharedSpaceAction(result)));
      store.dispatch(SharedSpaceDetailsGetAllSharedSpaceMembersAction(
          await _getAllSharedSpaceMembersInteractor.execute(sharedSpaceId)));
    });
  }
```

hoangdat:
> > I really want state for `add_member` also. Idealy, we should use `state` as much as possible. Only one thing to care in the app, it is `appState`. It is easier for us when debugging instead of finding somewhere to store something. One and only one: `AppState`. It is the point why we use Redux. @imGok @huynguyennovem @dab246 @manhduc-nguyen
> 
> 
> 
> For me, Redux is used to store persistent data across the app not view data (volatile data).
> 
> I don't want to use state for everything since onChanges listen to all state, so if we charge state more it will slow the app.
> 
> In our case, selectedRole is used only in this view.
> 
> 
> 
> I can try to find a solution with a stream builder tho

But for me, I dont want the view is rebuild by any other cause except app state.

lib/presentation/widget/shared_space_details/add_shared_space_member/add_shared_space_member_viewmodel.dart
```diff
@@ -0,0 +1,132 @@
...
+      SharedSpaceId sharedSpaceId, SharedSpaceMemberAutoCompleteResult autoCompleteResult) {
+    store.dispatch(
+        _addSharedSpaceAction(sharedSpaceId, AccountId(autoCompleteResult.userUuid.uuid)));
+  }
+
+  OnlineThunkAction _addSharedSpaceAction(SharedSpaceId sharedSpaceId, AccountId accountId) {
+    return OnlineThunkAction((Store<AppState> store) async {
+      final rolesList = store.state.sharedSpaceState.rolesList;
+      if (rolesList.isEmpty) {
+        return;
```
hoangdat, line 106: We can not response nothing for the user.

imGok:
@hoangdat check again pls

imGok:
need to fix toast tho

hoangdat (commented):
please squash it, quite good

imGok:
@hoangdat done

huynguyennovem (approved):

hoangdat (approved):

PR #296, opened 2021-03-31
Add Ru translation

manhduc-nguyen:


lib/l10n/intl_ru.arb
```diff
@@ -1,6 +1,6 @@
 {
   "@@last_modified": "2020-11-10T17:18:26.612601",
-  "initializing_data": "Initializing data...",
+  "initializing_data": "Инициализация данных ...",
```
huynguyennovem, line 3: The three-dot here is different than this one `Подготовка к загрузке…` ?

manhduc-nguyen: oops mb

imGok:
Looks good to me (can't verify ^^)

imGok (approved):

hoangdat (approved):

PR #325, opened 2021-04-12
Some bugs related to workgroup on latest test

hoangdat:


lib/presentation/util/widget_utils.dart
```diff
@@ -0,0 +1,38 @@
...
+ * applicable Additional Terms for LinShare along with this program. If not, see
+ * <http://www.gnu.org/licenses/> for the GNU Affero General Public License version
+ *  3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf> for
+ *  the Additional Terms applicable to LinShare software.
+ */
+
+import 'package:flutter/widgets.dart';
+
+class WidgetConstant {
+  static final Widget NO_WIDGET = null;
```
dab246, line 37: I don't think it is necessary to assign null

dab246 (commented):
LGTM

lib/presentation/util/widget_utils.dart
hoangdat: It is readable :)

huynguyennovem (commented):
/lgtm

PR #347, opened 2021-04-20
LA-231 Search in received shares

imGok:
search


hoangdat:
quite good, but the padding of the search input is not equally. Please squash and fix up this. @imGok

hoangdat (approved):

PR #379, opened 2021-04-29
Bug_fixes_align_header_of_context_menu

datnguyenhuy:
image
image
image
image

### DIFF
--- a/lib/presentation/view/header/context_menu_header_builder.dart
+++ b/lib/presentation/view/header/context_menu_header_builder.dart
@@ -32,6 +32,8 @@
 import 'package:filesize/filesize.dart';
 import 'package:flutter/material.dart';
 import 'package:linshare_flutter_app/presentation/model/file/presentation_file.dart';
+import 'package:linshare_flutter_app/presentation/model/file/shared_space_node_nested_presentation_file.dart';
+import 'package:linshare_flutter_app/presentation/model/file/work_group_folder_presentation_file.dart';
 import 'package:linshare_flutter_app/presentation/util/extensions/color_extension.dart';
 
 class ContextMenuHeaderBuilder {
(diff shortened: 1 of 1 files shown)


lib/presentation/model/file/document_presentation_file.dart
```diff
@@ -102,6 +102,7 @@ class DocumentPresentationFile extends Equatable implements PresentationFile {
 
   @override
   Widget fileIcon() {
+    print(mediaType.getFileTypeImagePath(imagePath));
```
hoangdat, line 105: please remove that

lib/presentation/model/file/presentation_file.dart
```diff
@@ -42,4 +42,6 @@ abstract class PresentationFile {
   int fileSize();
 
   Widget fileIcon();
+
+  String fileIconImagePath();
```
hoangdat, line 46: why we need this method?

lib/presentation/view/header/context_menu_header_builder.dart
```diff
@@ -40,18 +40,21 @@ class ContextMenuHeaderBuilder {
...
+        child: ListTile(
+          key: _key,
+          leading: Transform.translate(
+              offset: _file.fileIconImagePath() == 'assets/images/ic_folder.svg' ? Offset(0,0) : Offset(0,2),
+              child: _file.fileIcon()
           ),
-        ),
-      trailing: _file.fileSize() == 0 ? null : Text(filesize(_file.fileSize()), style: TextStyle(fontSize: 14, color: AppColor.documentModifiedDateItemTextColor)));
+          title: Text(
+            _file.fileName(),
```
hoangdat, line 53: How about if `fileName` is too long?

lib/presentation/view/header/context_menu_header_builder.dart
```diff
@@ -40,18 +40,21 @@ class ContextMenuHeaderBuilder {
...
-          _file.fileName(),
-          maxLines: 1,
-          style: TextStyle(fontSize: 14, color: AppColor.documentNameItemTextColor),
+  Transform build() {
+    return Transform.translate(
+        offset: Offset(0,3.5),
+        child: ListTile(
+          key: _key,
+          leading: Transform.translate(
+              offset: _file.fileIconImagePath() == 'assets/images/ic_folder.svg' ? Offset(0,0) : Offset(0,2),
```
imGok, line 49: Use constants instead please and I think you `fileIconImagePath` is not needed

imGok: You could do : `_file is WorkGroupFolderPresentationFile ? ..` or smth like this wdyt ?

datnguyenhuy: It was so good that I couldn't think of it. Thank you very much

lib/presentation/model/file/presentation_file.dart
datnguyenhuy: It is not needed and i removed it. Thank you, please

lib/presentation/view/header/context_menu_header_builder.dart
datnguyenhuy: I added the ellipsis at the end if it's too long

lib/presentation/view/header/context_menu_header_builder.dart
```diff
@@ -40,18 +42,22 @@ class ContextMenuHeaderBuilder {
...
-    return ListTile(
-      key: _key,
-      leading: _file.fileIcon(),
-      title: Transform(
-        transform: Matrix4.translationValues(-16, -2, 0.0),
-        child: Text(
-          _file.fileName(),
-          maxLines: 1,
-          style: TextStyle(fontSize: 14, color: AppColor.documentNameItemTextColor),
+  Transform build() {
```
hoangdat, line 45: Can you explain the different from List tile with Transform

datnguyenhuy: ListTile is used to display the list box
Transform is used to align ListTile on the header context menu, taking ListTile as a child widget

hoangdat (approved):

PR #383, opened 2021-04-29
LA-321 Wrong message cancel Preview

imGok:
cancelPreview

Fixed in every views


domain/lib/src/usecases/download_file/download_file_exception.dart
```diff
@@ -29,18 +29,34 @@
 //  3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf> for
 //  the Additional Terms applicable to LinShare software.
 
-class DeviceNotSupportedException implements Exception {
-  String notSupported = 'This device is not supported, please try on Android or iOS';
-}
+import 'package:equatable/equatable.dart';
 
-class DownloadFileException implements Exception {
+abstract class DownloadFileException extends Equatable implements Exception {
```
hoangdat, line 34: why not `with Equaltable`

domain/lib/src/usecases/download_file/download_file_exception.dart
```diff
@@ -29,18 +29,34 @@
...
+
+class CancelDownloadFileException extends DownloadFileException {
+  CancelDownloadFileException(cancelMessage) : super(cancelMessage);
+
+  @override
+  List<Object> get props => [message];
+}
 
-  CancelDownloadFileException(this.message);
+class DeviceNotSupportedException extends DownloadFileException {
```
hoangdat, line 60: where you use this exception?

hoangdat (commented):
It quite good

domain/lib/src/usecases/download_file/download_file_exception.dart
imGok: I don't know to be honest I'm not familiar with mixin so that's why I used extends

domain/lib/src/usecases/download_file/download_file_exception.dart
imGok: I just refactored extensions so they can be equatable, this one already existed before

dab246 (approved):
good

hoangdat (approved):

PR #387, opened 2021-05-04
LA-371 Apply tablet design for workgroup

dab246:
https://user-images.githubusercontent.com/80730648/116990855-e5f9f880-acfd-11eb-962e-d40aafcddb04.mov

### DIFF
--- a/lib/presentation/util/helper/responsive_widget.dart
+++ b/lib/presentation/util/helper/responsive_widget.dart
@@ -50,7 +50,7 @@ class ResponsiveWidget extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    var widthDevice = MediaQuery.of(context).size.width;
+    var widthDevice = MediaQuery.of(context).size.shortestSide;
     return LayoutBuilder(builder: (context, constraints) {
       if (widthDevice >= minLargeWidth) {
         return largeScreen ?? mediumScreen;
--- a/lib/presentation/util/helper/responsive_widget.dart
+++ b/lib/presentation/util/helper/responsive_widget.dart
@@ -36,8 +36,10 @@ class ResponsiveWidget extends StatelessWidget {
   final Widget mediumScreen;
   final Widget smallScreen;
 
-  static final int minLargeWidth = 960;
-  static final int minMediumWidth = 640;
+  static final int minLargeWidth = 950;
+  static final int minMediumWidth = 600;
+
+  static final double horizontalPaddingLargeWidth = 132.0;
 
   const ResponsiveWidget({
     Key key,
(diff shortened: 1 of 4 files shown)


lib/presentation/util/helper/responsive_widget.dart
```diff
@@ -36,8 +36,10 @@ class ResponsiveWidget extends StatelessWidget {
   final Widget mediumScreen;
   final Widget smallScreen;
 
-  static final int minLargeWidth = 960;
-  static final int minMediumWidth = 640;
+  static final int minLargeWidth = 950;
+  static final int minMediumWidth = 600;
```
hoangdat, line 40: why we need to change this?

lib/presentation/util/helper/responsive_widget.dart
```diff
@@ -36,8 +36,10 @@ class ResponsiveWidget extends StatelessWidget {
...
+
+  static final double horizontalPaddingLargeWidth = 132.0;
```
hoangdat, line 42: can you explain to me?

lib/presentation/view/order_by/order_by_button.dart
```diff
@@ -60,16 +61,18 @@ class OrderByButtonBuilder {
...
-          _currentSorter.orderType.getSortIconPath(imagePath),
-          width: 15,
-          height: 15,
-          fit: BoxFit.fill,
-          color: AppColor.primaryColor,
-        ),
-      ),
+      contentPadding: ResponsiveWidget.isLargeScreen(_context)
+          ? EdgeInsets.symmetric(horizontal: 155)
+          : EdgeInsets.symmetric(horizontal: 16) ,
```
hoangdat, line 66: srr, I can not understand

lib/presentation/util/helper/responsive_widget.dart
dab246: For a large tablet screen by design, we need to indent the listview items in the middle. So I set the value to all widgets that I need to use.

lib/presentation/util/helper/responsive_widget.dart
dab246: These are the two most common values for large and xlarge tablet screens. You can see more here. [https://developer.android.com/training/multiscreen/screensizes](https://developer.android.com/training/multiscreen/screensizes)

lib/presentation/view/order_by/order_by_button.dart
dab246: We need a distance from the left margin on the large tablet screen

lib/presentation/util/helper/responsive_widget.dart
hoangdat: what is the unit for this width? (pixel? ...)

hoangdat: and IMO 950 is not a common value

dab246: it is dp

dab246: I have read through the documentation it is a popular desktop value. I don't remember the link of that document. Maybe we will discuss this value again.

hoangdat: iOS using dp too?

dab246: In flutter we use the same px. If dp is large then dp ~ px

PR #389, opened 2021-05-05
Cherry functions before biometric

hoangdat:
- List all versions
- Setting for biometric
- [Fix] Align search area


dab246 (approved):
Good

PR #396, opened 2021-05-07
Config iOS project for migration to null safety

manhduc-nguyen:


hoangdat:
@huynguyennovem can we close it?

huynguyennovem:
> @huynguyennovem can we close it?

yes

PR #412, opened 2021-05-12
Fix bug on master branch for android

dab246:
**BUG FOUND ON ANDROID**

- App crash when perform share action in context menu
- Repeat toast display in SharedSpace
- Dialog loading still displays when preview file return failed
- Rename workgroup failed

### DIFF
--- a/data/lib/src/network/linshare_http_client.dart
+++ b/data/lib/src/network/linshare_http_client.dart
@@ -493,9 +493,6 @@ class LinShareHttpClient {
     final resultJson = await _dioClient.put(
       Endpoint.sharedSpaces.withPathParameter(sharedSpaceId.uuid).generateEndpointPath(),
       data: renameWorkGroupBodyRequest.toJson().toString(),
-      options: Options(headers: {
-        'Content-Type': 'application/json; charset=UTF-8'
-      })
     );
 
     return SharedSpaceNodeNestedResponse.fromJson(resultJson);
--- a/domain/lib/src/extension/string_extension.dart
+++ b/domain/lib/src/extension/string_extension.dart
@@ -42,7 +42,13 @@ extension StringExtension on String {
   }
 
   String toMiddleEllipsis() {
-    final subLength = length - split('.').last.length;
-    return '...' + substring(subLength - 3, subLength) + split('.').last;
+    if (contains('.')) {
+      final subLength = length - split('.').last.length;
+      return subLength >= 3
+          ? '...' + substring(subLength - 3, subLength) + split('.').last
+          : this;
+    } else {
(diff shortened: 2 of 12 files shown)


domain/lib/src/model/sharedspace/versioning_parameter.dart
```diff
@@ -37,7 +37,7 @@ class VersioningParameter extends Equatable {
   VersioningParameter(this.enable);
 
   factory VersioningParameter.initial() {
-    return VersioningParameter(false);
+    return VersioningParameter(true);
```
hoangdat, line 40: why we need this change?

hoangdat (commented):
@huynguyennovem please review this

domain/lib/src/model/sharedspace/versioning_parameter.dart
dab246: That is the default value when the returned shared space object has no VersioningParameter value. I checked the web version and found that the initial setting is like that. I also don't understand why they put it like that for this object. Even so, I have checked that setting `true` or `false` has no effect

domain/lib/src/extension/string_extension.dart
```diff
@@ -43,6 +43,8 @@ extension StringExtension on String {
 
   String toMiddleEllipsis() {
     final subLength = length - split('.').last.length;
-    return '...' + substring(subLength - 3, subLength) + split('.').last;
```
huynguyennovem, line 46: Some file may have no extension, so I think we should check more cases, like this:

```dart
if(contains('.')) {
	final subLength = length - split('.').last.length;
	return subLength >= 3
	  ? '...' + substring(subLength - 3, subLength) + split('.').last
	  : this;
} else {
  	return length >= 3 ? '...' + substring(length - 3, length) : this;
}
```

Or to use package https://pub.dev/packages/path to check extension exists

data/lib/src/network/linshare_http_client.dart
```diff
@@ -493,9 +493,6 @@ class LinShareHttpClient {
     final resultJson = await _dioClient.put(
       Endpoint.sharedSpaces.withPathParameter(sharedSpaceId.uuid).generateEndpointPath(),
       data: renameWorkGroupBodyRequest.toJson().toString(),
-      options: Options(headers: {
-        'Content-Type': 'application/json; charset=UTF-8'
-      })
```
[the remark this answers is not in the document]
dab246: Not because of it. Because it misses the nodeType parameter on v4.0 only. I have checked `Content-Type` here which is not necessary because it is already set for us by default. So I deleted it to avoid misunderstandings.

hoangdat (approved):

Issue #255, opened 2021-03-16
search term should be cleared after user Copy to Workgroup from search result


hantt12:
#### Reproduction Steps
1. User logs in and open MySpace
2. User clicks on Search icon
3. User inputs a search term to search box 
4. In the search result list, view list of actions for a file
5. Select "Copy to a workgroup"
6. Search a workgroup and click Copy Here

### Expected result
Show all files in MySpace

### Current behavior
Show list of files as search result (but without search box)

### Context

Iphone XR
Linshare 0.4.0 (36)

hoangdat:
@dieptran88 please verify this issue?

dieptran88:
I confirmed the current behavior,
Expected result: Show the previous screen which is search result screen with search box

hoangdat:
v2.0.0-alpha02

PR #414, opened 2021-05-12
LA-255 Clear Search View After Copying

imGok:
#255 

clearsearch


imGok:
Basically, "TextEditingController.clear()" is needed to clear the search bar and it was a stateless widget so now we can clear the search term outside the widget

hoangdat:
Can you update this PR?

hoangdat:
make it work with `features_v2`

hoangdat:
@huynguyennovem please review this PR also

lib/presentation/di/module/widget_module.dart
```diff
@@ -446,4 +449,10 @@ class WidgetModule {
         ));
   }
 
+  void _provideHomeAppBarComponent() {
+    getIt.registerFactory(() => HomeAppBarWidget());
```
huynguyennovem, line 453: Registered `HomeAppBarWidget` but there is no usage of it with `getIt`

imGok: you are right I just need the viewmodel

hoangdat (approved):

PR #427, opened 2021-05-15
bug fixes Delete dialog overflows to the screen when the file name is too long

datnguyenhuy:
image
_title is one line long
image
_title is two lines long
image
_title is three lines long


huynguyennovem:
I am not sure this suitable solution to fix overflow. Let's try with 4 or more lines. IMO, the title should have 2 lines maximum and have ellipsis text too.

dab246:
> I am not sure this suitable solution to fix overflow. Let's try with 4 or more lines. IMO, the title should have 2 lines maximum and have ellipsis text too.

I don't think it can overflow. Because he has limited `maxLines = 3` already.  
Screen Shot 2021-05-18 at 17 28 11

dab246 (approved):
LGTM

huynguyennovem:
I think the root cause from this: https://github.com/linagora/linshare-mobile-flutter-app/blob/ab9152d3e6c25f41fce6e6e72a2f2961deaa4491/lib/presentation/view/modal_sheets/confirm_modal_sheet_builder.dart#L91

PR #446, opened 2021-05-21
Fix bug: Preview OpenOffice file type

huynguyennovem:
- Bug description:
Can not preview OpenOffice file type, such as: .odt, .ods

- Demo (after fixing):

https://user-images.githubusercontent.com/29337364/119128409-d39eef00-ba5f-11eb-91ab-6148260bb4fc.mp4


hoangdat:
with this fix, How about in iOS, I guess we can not see the open-document in iOS. @huynguyennovem

PR #448, opened 2021-05-24
Hotfix: overflow when list recipients too long

huynguyennovem:


hoangdat:
can you show a screenshot?

huynguyennovem:
@hoangdat  sr, i missed it. It is:
Simulator Screen Shot - iPhone 12 Pro Max - 2021-05-24 at 12 45 11

hoangdat:
hmm. `Share` button overlapped the recipient list. How about make the recipient list above this button?

hoangdat (approved):

PR #466, opened 2021-05-25
LA-462,463 Fix bug destination picker

dab246:
**LA-462 Fix bug blank screen when copy files from Workgroup to MySpace in offline mode**

https://user-images.githubusercontent.com/80730648/119472734-afe7eb80-bd74-11eb-8de9-b744db9eeddd.mov

**LA-463 Fix bug pick destination screen is not shown in landscape view**

https://user-images.githubusercontent.com/80730648/119472827-c68e4280-bd74-11eb-87b1-8b764fa9256e.mov

### DIFF
--- a/lib/presentation/util/helper/responsive_utils.dart
+++ b/lib/presentation/util/helper/responsive_utils.dart
@@ -53,6 +53,8 @@ class ResponsiveUtils {
 
   static const double _loginButtonWidth = 240.0;
 
+  static const double _radiusDestinationPickerView = 20.0;
+
   double getSizeWidthScreen(BuildContext context) {
     return MediaQuery.of(context).size.width;
   }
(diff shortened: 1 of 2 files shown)


lib/presentation/util/helper/responsive_utils.dart
```diff
@@ -86,18 +88,44 @@ class ResponsiveUtils {
...
+            horizontal: _destinationPickerHorizontalMarginMediumScreen,
+            vertical: 0.0)
+        : EdgeInsets.symmetric(
+            horizontal: _destinationPickerHorizontalMarginMediumScreen,
+            vertical: _destinationPickerVerticalMarginMediumScreen);
     } else if (isLargeScreen(context)) {
-      return EdgeInsets.symmetric(
-        horizontal: _destinationPickerHorizontalMarginLargeScreen,
-        vertical: _destinationPickerVerticalMarginLargeScreen);
+      return getSizeHeightScreen(context) <= _destinationPickerVerticalMarginLargeScreen * 2
```
hoangdat, line 99: srr, but what is it `<=`?

lib/presentation/util/helper/responsive_utils.dart
```diff
@@ -86,18 +88,44 @@ class ResponsiveUtils {
...
 
+  BorderRadius getBorderRadiusView(BuildContext context) {
+    if (isMediumScreen(context)) {
+      return getSizeHeightScreen(context) <= _destinationPickerVerticalMarginMediumScreen * 2
+        ? BorderRadius.only(
+            topLeft: Radius.circular(_radiusDestinationPickerView),
+            topRight: Radius.circular(_radiusDestinationPickerView))
+        : BorderRadius.circular(_radiusDestinationPickerView);
+    } else if (isLargeScreen(context)) {
+      return getSizeHeightScreen(context) <= _destinationPickerVerticalMarginLargeScreen * 2
```
hoangdat, line 119: Can you explain a little bit?

lib/presentation/util/helper/responsive_utils.dart
dab246: size height screen

lib/presentation/util/helper/responsive_utils.dart
dab246: We check when the height of the screen is less than or equal to the distance to the margin that we have set when displaying on tablet devices, we will not be separated from the margin vertically but only horizontally.

hoangdat (approved):

PR #467, opened 2021-05-25
LA-434 Fix grammar mistakes

imGok:


lib/presentation/localizations/app_localizations.dart
```diff
@@ -1022,154 +1022,308 @@ class AppLocalizations {
...
-  String audit_action_message_create_workgroup(String authorName, String resourceName) {
+  String audit_action_message_create_workgroup_self(String resourceName) {
+    return Intl.message(
+        'You have created a new Workgroup : $resourceName.',
+        name: 'audit_action_message_create_workgroup_self',
+        args: [resourceName]
+    );
+  }
+
+  String audit_action_message_create_workgroup_other(String authorName, String resourceName) {
```
hoangdat, line 1033: we use it in where?

lib/l10n/intl_en.arb
```diff
@@ -756,168 +756,286 @@
...
-  "audit_action_message_update_workgroup": "{authorName} has edited the Workgroup : {resourceName}.",
-  "@audit_action_message_update_workgroup": {
+  "audit_action_message_delete_workgroup_self": "You have deleted the Workgroup : {resourceName}.",
+  "@audit_action_message_delete_workgroup_self": {
+    "type": "text",
+    "placeholders": {
+      "resourceName": {}
+    }
+  },
+  "audit_action_message_delete_workgroup_other": "{authorName} has deleted the Workgroup : {resourceName}.",
```
hoangdat, line 781: same for it, and `other` string

lib/l10n/intl_en.arb
```diff
@@ -1262,54 +1370,6 @@
   },
   "some_files_could_not_be_duplicated": "Some files could not be duplicated",
   "@some_files_could_not_be_duplicated": {
-    "type": "text",
-    "placeholders": {
-      "biometricKind": {}
-    }
-  },
-  "biometric_disabled_in_setting_app": "Please enable {biometricKind} in the Settings app on your phone to use biometric authentication",
```
hoangdat, line 1270: why we delete that?

lib/l10n/intl_en.arb
```diff
@@ -756,168 +756,286 @@
     "type": "text",
     "placeholders": {}
   },
-  "audit_action_message_create_workgroup": "{authorName} has created a new Workgroup : {resourceName}.",
-  "@audit_action_message_create_workgroup": {
+  "audit_action_message_create_workgroup_self": "You have created a new Workgroup : {resourceName}.",
```
dab246, line 759: Why did you fix it like that? What if the `authorName ` passed in is sometimes not `You`. You should double check my code flow before fixing it. Be careful, errors will occur.

dab246 (commented):
You need to double check this feature when making such a change. !!!

PR #474, opened 2021-05-27
LA-458 Get Upload Request Groups [skip-ci]

imGok:
uploadrequests

PR #486, opened 2021-06-07
LA-469 View upload requests of a group

imGok:
uploadrequestsList


imGok:
I'll add Int strings at the end (easier to rebase)

lib/presentation/widget/upload_request/upload_request_viewmodel.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'package:linshare_flutter_app/presentation/widget/base/base_viewmodel.dart';
+import 'package:redux/redux.dart';
+
+class UploadRequestViewModel extends BaseViewModel {
+  final AppNavigation _appNavigation;
+  final GetAllUploadRequestInteractor _getAllUploadRequestInteractor;
+
+  UploadRequestViewModel(
+      Store<AppState> store, this._appNavigation, this._getAllUploadRequestInteractor)
+      : super(store);
```
dab246, line 46: Indent should be 2 tabs

lib/presentation/widget/upload_request/upload_request_widget.dart
```diff
@@ -0,0 +1,226 @@
...
+            width: 24,
+            height: 24,
+            fit: BoxFit.fill,
+          ),
+          onPressed: () => null),
+    );
+  }
+
+  Widget _buildActiveClosedSubtitleWidget(UploadRequestStatus status) {
+    if (status == UploadRequestStatus.ENABLED) {
```
dab246, line 123: You can use the `Switch-case` here.

lib/presentation/redux/states/upload_request_state.dart
```diff
@@ -0,0 +1,77 @@
...
+  @override
+  UploadRequestState startLoadingState() {
+    return UploadRequestState(Right(LoadingState()), uploadRequestsList);
+  }
+
+  @override
+  List<Object> get props => [
+        ...super.props,
+        uploadRequestsList,
+      ];
```
dab246, line 76: Indent should be 2 tabs

lib/presentation/redux/actions/upload_request_action.dart
```diff
@@ -0,0 +1,58 @@
...
+import 'package:dartz/dartz.dart';
+import 'package:domain/domain.dart';
+import 'package:flutter/foundation.dart';
+import 'package:linshare_flutter_app/presentation/redux/actions/app_action.dart';
+
+@immutable
+class StartUploadRequestLoadingAction extends ActionOnline {}
+
+@immutable
+class UploadRequestAction extends ActionOffline {
```
dab246, line 42: I think it's `online` right

imGok: It's just an action to change the view state so it can be both online and offline

data/lib/src/network/model/response/upload_request_response.dart
```diff
@@ -0,0 +1,172 @@
...
+      recipients.map((recipient) => recipient.toUploadRequestAccount()).toSet(),
+      activationDate,
+      modificationDate,
+      creationDate,
+      expiryDate,
+      notificationDate,
+      label,
+      status,
+      maxFileCount,
+      maxDepositSize,
```
huynguyennovem, line 158: Because it could be null, it should have an unexpected/default value here. Such as: `maxDepositSize ?? 0`. The same to others

lib/presentation/widget/upload_request_group/upload_request_group_widget.dart
```diff
@@ -0,0 +1,212 @@
...
+            width: 24,
+            height: 24,
+            fit: BoxFit.fill,
+          ),
+          onPressed: () => null),
+        onTap: () => _model.goToUploadRequests(request),
+    );
+  }
+
+  Widget _buildActiveClosedSubtitleWidget(UploadRequestStatus status) {
```
huynguyennovem, line 168: It's duplicated LOC for these cases. They are the only differences in the Decoration color. Pls write a common function for them.

lib/presentation/widget/upload_request/upload_request_widget.dart
```diff
@@ -0,0 +1,226 @@
...
+            imagePath.icContextMenu,
+            width: 24,
+            height: 24,
+            fit: BoxFit.fill,
+          ),
+          onPressed: () => null),
+    );
+  }
+
+  Widget _buildActiveClosedSubtitleWidget(UploadRequestStatus status) {
```
huynguyennovem, line 122: Let's write the common function for these cases. There are duplicated LOC

huynguyennovem:
Pls also add the prefix `(DO NOT REVIEW)` before the commits that come from PR #474

lib/presentation/widget/upload_request_group/upload_request_group_widget.dart
imGok: Text are also differents between status so I think separate them is a good thing

data/lib/src/network/model/response/upload_request_response.dart
imGok: You are right, what default value I should put for dateTime ?

huynguyennovem: I think it should be a constant DateTime, a special time, such as: 01/01/2001. It may help us to debug faster afterward

lib/presentation/widget/upload_request_group/upload_request_group_widget.dart
huynguyennovem: I mean this code block:

```
return Row(children: [
  Container(
    margin: EdgeInsets.only(right: 4),
    width: 7.0,
    height: 7.0,
    decoration: BoxDecoration(
      color: AppColor.primaryColor,
      shape: BoxShape.circle,
    ),
  ),
  Text(AppLocalizations.of(context).active,
      style: TextStyle(fontSize: 14, color: AppColor.uploadFileFileSizeTextColor))
]);
```

Then only passing the different things via params

PR #490, opened 2021-06-08
LA-? Received Shares Details

imGok:


lib/presentation/widget/received/received_share_viewmodel.dart
```diff
@@ -388,6 +390,12 @@ class ReceivedShareViewModel extends BaseViewModel {
     store.dispatch((ReceivedShareSetSearchResultAction([])));
   }
 
+  void goToDetails(ReceivedShare receivedShare) {
+    _appNavigation.popBack();
```
huynguyennovem, line 394: why don't you use `popAndPush` instead?

lib/presentation/widget/received_share_details/received_share_details_widget.dart
```diff
@@ -0,0 +1,205 @@
...
+              height: 30,
+              child: CircularProgressIndicator(
+                backgroundColor: AppColor.primaryColor,
+              ),
+            ),
+          ));
+    }
+
+    return SingleChildScrollView(
+        child: Column(
```
huynguyennovem, line 120: pls format to have the same indent level of SingleChildScrollView's params

lib/presentation/widget/received/received_share_viewmodel.dart
imGok: you are totally right ^^

hoangdat (approved):

PR #546, opened 2021-07-15
Add Intl Utils

imGok:
Since `flutter pub run intl_translation:generate_from_arb` doesn't generate files with null safety, I suggest to add a recent package that support null safety : https://pub.dev/packages/intl_utils 

command : `flutter pub run intl_utils:generate`

We can discuss about output file tho

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -118,6 +118,7 @@ dependencies:
 
   # OIDC
   flutter_appauth: 1.0.0+1
+  intl_utils: ^2.4.0
 
   # datetime picker
   flutter_datetime_picker: 1.5.1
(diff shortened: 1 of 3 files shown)


pubspec.yaml
```diff
@@ -181,3 +182,23 @@ flutter:
   #
   # For details regarding fonts from package dependencies,
   # see https://flutter.dev/custom-fonts/#from-packages
+
```
hoangdat, line 185: it a big change, please put an ADR to clear about that. Thanks

huynguyennovem:
There is another simpler solution for this:

Add this in the end of `pubspec.yaml` file:
```
flutter_intl:
  enabled: true
```

hoangdat:
@imGok any update for this.

imGok:
@hoangdat I still want to implement this because the file crashed all the time

pubspec.yaml
hoangdat: @imGok please describe your solution in ADR.

hoangdat:
@imGok @huynguyennovem IMO, we don't need this fix.

imGok:
Fixed by adding intl_generator

PR #547, opened 2021-07-21
LA-534 Create Upload Request (Simple)

huynguyennovem:
https://user-images.githubusercontent.com/29337364/126636212-e6c39347-3cbf-419a-b782-33bbb0706573.mov

### DIFF
--- a/data/lib/src/datasource_impl/upload_request_group_datasource_impl.dart
+++ b/data/lib/src/datasource_impl/upload_request_group_datasource_impl.dart
@@ -59,4 +59,20 @@ class UploadRequestGroupDataSourceImpl implements UploadRequestGroupDataSource {
       });
     });
   }
+
+  @override
+  Future<UploadRequestGroup> addNewUploadRequest(UploadRequestCreationType creationType, AddUploadRequest addUploadRequest) async {
+    return Future.sync(() async {
+      final uploadRequestGroupResponse = await _linShareHttpClient.addNewUploadRequest(creationType, addUploadRequest);
+      return uploadRequestGroupResponse.toUploadRequestGroup();
+    }).catchError((error) {
+      _remoteExceptionThrower.throwRemoteException(error, handler: (DioError error) {
+        if (error.response?.statusCode == 404) {
+          throw UploadRequestCreateFailed();
+        } else {
+          throw UnknownError(error.response?.statusMessage);
+        }
+      });
+    });
+  }
 }
--- a/data/lib/src/network/linshare_http_client.dart
+++ b/data/lib/src/network/linshare_http_client.dart
@@ -67,6 +67,7 @@ import 'package:domain/domain.dart';
 
 import 'model/autocomplete/shared_space_member_autocomplete_result_dto.dart';
 import 'model/request/add_shared_space_member_body_request.dart';
+import 'model/request/add_upload_body_request.dart';
 import 'model/request/copy_body_request.dart';
 import 'model/request/create_shared_space_node_folder_request.dart';
 import 'model/request/update_shared_space_member_body_request.dart';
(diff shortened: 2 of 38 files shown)


data/lib/src/datasource_impl/upload_request_group_datasource_impl.dart
```diff
@@ -59,4 +59,20 @@ class UploadRequestGroupDataSourceImpl implements UploadRequestGroupDataSource {
...
+
+  @override
+  Future<UploadRequestGroup> addNewUploadRequest(UploadRequestCreationType creationType, AddUploadRequest addUploadRequest) async {
+    return Future.sync(() async {
+      final uploadRequestGroupResponse = await _linShareHttpClient.addNewUploadRequest(creationType, addUploadRequest);
+      return uploadRequestGroupResponse.toUploadRequestGroup();
+    }).catchError((error) {
+      _remoteExceptionThrower.throwRemoteException(error, handler: (DioError error) {
+        if (error.response?.statusCode == 404) {
+          throw UploadRequestGroupsNotFound();
```
hoangdat, line 71: it really response this error

data/lib/src/network/linshare_http_client.dart
```diff
@@ -546,4 +547,14 @@ class LinShareHttpClient {
 
     return resultJson.map((data) => UploadRequestGroupResponse.fromJson(data)).toList();
   }
+
+  Future<UploadRequestGroupResponse> addNewUploadRequest(UploadRequestCreationType creationType, AddUploadRequest addUploadRequest) async {
+    final resultJson = await _dioClient.put(
```
hoangdat, line 552: IMO, it should be `POST` method

lib/presentation/util/extensions/datetime_extension.dart
```diff
@@ -35,4 +35,9 @@ import 'package:intl/intl.dart';
 extension DateTimeExtension on DateTime {
 
   String getMMMddyyyyFormatString() => DateFormat.yMMMd().format(this);
+
+  String get_yMMMMd_formatWithJm() => DateFormat.yMMMMd().add_jm().format(this);
```
hoangdat, line 39: What is `jm`? Besides that, IMO, you should not use this format for your method name?

hoangdat (commented):
Too many things in one commit, please try to split as small as possible next time. We can split the presentation commit into some components:
- add users
- input messages
- other data
- ...
Very hard to get your idea in your solution.

data/lib/src/network/model/request/add_upload_body_request.dart
```diff
@@ -75,7 +77,24 @@ class AddUploadBodyRequest with EquatableMixin {
 
   factory AddUploadBodyRequest.fromJson(Map<String, dynamic> json) => _$AddUploadBodyRequestFromJson(json);
 
-  Map<String, dynamic> toJson() => _$AddUploadBodyRequestToJson(this);
```
hoangdat: What problem with it?

lib/presentation/util/extensions/datetime_extension.dart
huynguyennovem: `add_jm()` used for time format

data/lib/src/network/model/request/add_upload_body_request.dart
huynguyennovem: The generation code causes an error when parsing the request object

lib/presentation/util/extensions/datetime_extension.dart
hoangdat: It is not answered my question? What is it? Why we need it?

huynguyennovem: It a function for time format. Example: `DateFormat.yMd().add_jm() -> 7/10/1996 5:08 PM`

hoangdat: No, we should not use this schema.

hoangdat (approved):

PR #551, opened 2021-07-23
LA-522 List files of Collective Upload Request

huynguyennovem:
https://user-images.githubusercontent.com/29337364/127185240-6bd345e2-71ce-4d13-8e9a-4d3660addab6.mov


data/lib/src/network/model/response/upload_request_entry_response.dart
```diff
@@ -0,0 +1,122 @@
...
+  final DateTime creationDate;
+  final DateTime modificationDate;
+  final DateTime expirationDate;
+  final String name;
+  final String comment;
+  final String uuid;
+  final String metaData;
+  final bool cmisSync;
+  final int size;
+  final String type;
```
hoangdat, line 56: Should be enum?

hoangdat (commented):
I did not see any test for repository

hoangdat:
please share the video demo also

lib/presentation/widget/upload_request_inside/upload_request_inside_navigator_widget.dart
```diff
@@ -0,0 +1,137 @@
...
+import 'package:linshare_flutter_app/presentation/widget/upload_request_inside/upload_request_document_type.dart';
+import 'package:linshare_flutter_app/presentation/widget/upload_request_inside/upload_request_inside_widget.dart';
+import 'package:rxdart/rxdart.dart';
+
+import '../../util/data_structure/stack.dart' as structure;
+
+typedef OnBackUploadRequestClickedCallback = void Function();
+typedef OnNodeClickedCallback = void Function(UploadRequestGroup uploadRequestGroup);
+
+class UploadRequestNavigatorWidget extends StatefulWidget {
```
hoangdat, line 46: Can you say sth about this widget. I still not have the idea why we need to maintain a stack?

lib/presentation/widget/upload_request_inside/upload_request_inside_navigator_widget.dart
```diff
@@ -0,0 +1,137 @@
...
+  @override
+  UploadRequestNavigatorWidgetState createState() => UploadRequestNavigatorWidgetState();
+}
+
+class UploadRequestNavigatorWidgetState extends State<UploadRequestNavigatorWidget> {
+
+  final GlobalKey<NavigatorState> navigatorKey = GlobalKey();
+
+  final structure.Stack<UploadRequestArguments> argumentStack = structure.Stack();
+  UploadRequestArguments? get currentPageData => argumentStack.peek();
```
imGok, line 71: Do you have any documents I can refer to for this kind of stack ? It's so interesting

imGok:
Nice code, overall there are few spaces / indentation mistakes but that's all I can say

lib/presentation/widget/upload_request_inside/upload_request_inside_navigator_widget.dart
huynguyennovem: Because we only need one widget for the Collective's files, Individual's recipients, Individual recipients files.. and beyond.  So, a stack is a suitable solution for kind of this.

lib/presentation/widget/upload_request_inside/upload_request_inside_navigator_widget.dart
huynguyennovem: It's only the normal stack. There is a library [here](https://pub.dev/packages/stack), but it's not necessary to use a library. We defined here: https://github.com/linagora/linshare-mobile-flutter-app/blob/master/lib/presentation/util/data_structure/stack.dart

hoangdat (approved):

PR #552, opened 2021-07-27
LA-513 Sort upload request

imGok:
sortuploadrequest

### DIFF
--- a/domain/lib/src/extension/list_upload_request_group_extension.dart
+++ b/domain/lib/src/extension/list_upload_request_group_extension.dart
@@ -28,12 +28,30 @@
 // <http://www.gnu.org/licenses/> for the GNU Affero General Public License version
 //  3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf> for
 //  the Additional Terms applicable to LinShare software.
-enum OrderScreen {
-  mySpace,
-  sharedSpaceDocument,
-  workGroup,
-  receivedShares,
-  uploadRequestGroupsCreated,
-  uploadRequestGroupsActiveClosed,
-  uploadRequestGroupsArchived
+
+import 'package:domain/domain.dart';
+
+extension ListUploadRequestGroup on List<UploadRequestGroup> {
+  void sortFiles(OrderBy orderBy, OrderType orderType) {
+    sort((file1, file2) {
+      switch (orderBy) {
+        case OrderBy.modificationDate:
+          return file2.modificationDate.compareToSort(file1.modificationDate, orderType);
+        case OrderBy.creationDate:
+          return file2.creationDate.compareToSort(file1.creationDate, orderType);
+        case OrderBy.expirationDate:
+          return file2.expiryDate.compareToSort(file1.expiryDate, orderType);
+        case OrderBy.activationDate:
+          return file2.activationDate.compareToSort(file1.activationDate, orderType);
+        case OrderBy.name:
+          return file2.label.compareToSort(file1.label, orderType);
+        case OrderBy.status:
+          return file2.status.compareToSort(file1.status, orderType);
+        case OrderBy.type:
+          return file2.collective.compareToSort(file1.collective, orderType);
+        default:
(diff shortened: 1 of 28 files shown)


huynguyennovem:
Manual test: 
- Missing order by: Status, Type
- Wrong one: `Name` (should be `Label`)
- After performing `Pull to refresh` (check the video I attached), the list must be kept as the recently Sorter (not the default)

https://user-images.githubusercontent.com/29337364/127439117-56547f29-041c-42a9-bfa2-83a80f7a1b99.mov

huynguyennovem:
I know that we has defined some sorters on the code base:
```dart
enum OrderBy {
  modificationDate,
  creationDate,
  expirationDate,
  activationDate,
  status,
  type,
  fileSize,
  name,
  shared,
  sender
}

...

case OrderBy.name:
        return AppLocalizations.of(context).name;

```

Adding a `label` seems not necessary in this case even though we're trying to make mobile application look like web version. I think we can keep it as `name` sorter. Can you help to confirm this? @hoangdat

domain/lib/src/extension/list_upload_request_group_extension.dart
```diff
@@ -0,0 +1,55 @@
...
+          return file2.modificationDate.compareToSort(file1.modificationDate, orderType);
+        case OrderBy.creationDate:
+          return file2.creationDate.compareToSort(file1.creationDate, orderType);
+        case OrderBy.expirationDate:
+          return file2.expiryDate.compareToSort(file1.expiryDate, orderType);
+        case OrderBy.activationDate:
+          return file2.activationDate.compareToSort(file1.activationDate, orderType);
+        case OrderBy.name:
+          return file2.label.compareToSort(file1.label, orderType);
+        // case OrderBy.status:
```
huynguyennovem, line 48: Missing Status and Type sorter

lib/presentation/widget/upload_request_group/upload_request_group_viewmodel.dart
```diff
@@ -47,10 +49,22 @@ import 'package:redux_thunk/redux_thunk.dart';
 class UploadRequestGroupViewModel extends BaseViewModel {
   final AppNavigation _appNavigation;
   final GetAllUploadRequestGroupsInteractor _getAllUploadRequestGroupsInteractor;
+  final GetSorterInteractor _getSorterInteractor;
```
huynguyennovem, line 52: Missing implementations for `_getSorterInteractor`

imGok: Yep forget it thx

domain/lib/src/extension/list_upload_request_group_extension.dart
imGok: Just added it, needed more informations

imGok:
@huynguyennovem pls review last commits :)

huynguyennovem:
> @huynguyennovem pls review last commits :)


https://user-images.githubusercontent.com/29337364/127636676-bb154be3-27c9-4f16-aeea-1c33385cc56f.mov

@imGok : Same Sorter but the list has changed after performing `Pull to refresh` action. Pls check other Sorters also.

huynguyennovem:
@imGok Look better but there is still a bug: The default sorter of empty list should be Modification date, not Name.

https://user-images.githubusercontent.com/29337364/127952457-587af44d-ea9a-4257-96c5-d8adcaa2a44d.mp4

imGok:
> @imGok Look better but there is still a bug: The default sorter of empty list should be Modification date, not Name.

I don't have this bug, are you sure you didn't put name before and return to the view ? because it saved the last sort

huynguyennovem:
> > @imGok Look better but there is still a bug: The default sorter of empty list should be Modification date, not Name.
> 
> I don't have this bug, are you sure you didn't put name before and return to the view ? because it saved the last sort

I'm not. This is the test video.

https://user-images.githubusercontent.com/29337364/128111362-a4fb9d05-6095-4be0-b41d-b5c06d75c05f.mov

Issue #560, opened 2021-08-03
[Bug] Back to wrong tab after go to Upload request


hoangdat:
### Description
Back to wrong tab after go to Upload request

### Expected result
Back to correctly Tab before go into an Upload Request

#### Reproduction Steps
1. Open Archived tab
2. Select 1 Upload Request
3. Back to Upload Request

### Acceptance criteria

The list of acceptance criteria (defined in the corresponding User Story or coming from your functional knowledge of the product)
These acceptance criteria must be met by the corrective implementation done by the developers.

### Context

Devices, OS

### Additional information

frequency, log files,...

PR #577, opened 2021-08-10
save last tab index in upload request group view (#560)

imGok:
sabtab

### DIFF
--- a/lib/presentation/widget/upload_request_group/upload_request_group_viewmodel.dart
+++ b/lib/presentation/widget/upload_request_group/upload_request_group_viewmodel.dart
@@ -67,6 +67,8 @@ class UploadRequestGroupViewModel extends BaseViewModel {
   SearchQuery _searchQuery = SearchQuery('');
   SearchQuery get searchQuery  => _searchQuery;
 
+  int _tabIndex = 0;
+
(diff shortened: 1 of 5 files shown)


lib/presentation/widget/upload_request_group/upload_request_group_viewmodel.dart
```diff
@@ -58,6 +58,8 @@ class UploadRequestGroupViewModel extends BaseViewModel {
   List<UploadRequestGroup> _uploadRequestListActiveClosed = [];
   List<UploadRequestGroup> _uploadRequestListArchived = [];
 
+  int _tabIndex = 0;
```
hoangdat, line 61: why we need store here? How about store it directly in UI state?

imGok: I thought dispatching an action at every tab switch was a bad idea and useless wdyt ?

hoangdat: yes, it should be consistent for all app. Not too many places to store the state of the application.

huynguyennovem (approved):
\lgtm

Issue #523, opened 2021-07-05
[Dev] Search in the list of files in Upload request 


huynguyennovem:
### Story
#453

### Desc:
- Collective: Upload request group -> list of files
- Individual: Upload request group -> list of recipients -> list of files

### DoD:
- Search for both files of Collective and Individual
- debounce the query string

![image.png](https://images.zenhubusercontent.com/5f7d8ef759d36aecc684e2a4/4a0f0e29-2b59-4007-9d96-83433bd0bb3c)

PR #608, opened 2021-08-31
Search Upload Request Inside (#523) 

imGok:
searchuploaddada

@hoangdat Can you verify the ui flow pls ?

- No search bar in Recipients list 
- Only in Upload Entries

### DIFF
--- a/domain/lib/src/usecases/search_upload_request_inside/search_upload_request_entries_interactor.dart
+++ b/domain/lib/src/usecases/search_upload_request_inside/search_upload_request_entries_interactor.dart
@@ -1,7 +1,7 @@
 // LinShare is an open source filesharing software, part of the LinPKI software
 // suite, developed by Linagora.
 //
-// Copyright (C) 2020 LINAGORA
+// Copyright (C) 2021 LINAGORA
 //
 // This program is free software: you can redistribute it and/or modify it under the
 // terms of the GNU Affero General Public License as published by the Free Software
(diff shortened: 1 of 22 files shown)


huynguyennovem:
Look good... Except this:

- When there is no item (entry) there, it should display text: "No files uploaded yet".  You have already reverted this by the [change](https://github.com/linagora/linshare-mobile-flutter-app/pull/608/files#diff-abc0dc7026b3b5f38948eb1df9d21b08dd92b841c3d05d0a058e7d77509f78d3R274) 

Screenshot_1630464554

- The original:
Screenshot_1630468674

huynguyennovem:
Look good... Except this:

- When there is no item (entry) there, it should display text: "No files uploaded yet".  You have already reverted this by the [change](https://github.com/linagora/linshare-mobile-flutter-app/pull/608/files#diff-abc0dc7026b3b5f38948eb1df9d21b08dd92b841c3d05d0a058e7d77509f78d3R274) 

Screenshot_1630464554

- The original:
Screenshot_1630468674

lib/presentation/localizations/app_localizations.dart
```diff
@@ -2160,6 +2160,12 @@ class AppLocalizations {
     );
   }
 
+  String get search_in_upload_requests {
```
huynguyennovem, line 2163: Should be "Search in Upload Request Entry". Because of we will have a "Search in Upload Request" for recipient (Upload Request) in the other tickets. Please also add string to all languages too

domain/lib/src/usecases/search_upload_request_inside/search_upload_request_entries_interactor.dart
```diff
@@ -0,0 +1,51 @@
...
+//  3 and <http://www.linshare.org/licenses/LinShare-License_AfferoGPL-v3.pdf> for
+//  the Additional Terms applicable to LinShare software.
+//
+
+import 'package:dartz/dartz.dart';
+import 'package:domain/domain.dart';
+
+class SearchUploadRequestEntriesInteractor {
+
+  Future<Either<Failure, Success>> execute(List<UploadRequestEntry> uploadRequestEntries, SearchQuery searchQuery) async {
```
huynguyennovem, line 38: Please add test case for this interactor

lib/presentation/widget/upload_request_inside/upload_request_inside_widget.dart
```diff
@@ -515,4 +537,27 @@ class _UploadRequestInsideWidgetState extends State<UploadRequestInsideWidget> {
       .build();
   }
 
+  // Search 
+  Widget _buildSearchViewWidget(BuildContext context) {
```
huynguyennovem, line 541: Please move this function to above all action functions. It will make us have consistent code for all widgets.

lib/presentation/redux/actions/upload_request_inside_action.dart
```diff
@@ -105,3 +105,15 @@ class UploadRequestSelectAllEntryAction extends ActionOffline {
 class UploadRequestUnSelectAllEntryAction extends ActionOffline {
   UploadRequestUnSelectAllEntryAction();
 }
+
+@immutable
+class UploadRequestSetSearchResultAction extends ActionOffline {
```
huynguyennovem, line 110: It should be `UploadRequestEntrySetSearchResultAction`

domain/lib/src/usecases/search_upload_request_inside/search_upload_request_entries_interactor.dart
imGok: Yes I did only functionnal code to check if feature is good, I usually do translations and tests at the end

imGok:
@huynguyennovem up to date !

domain/test/usecases/search_upload_request_entries/search_upload_request_entries_interactor_test.dart
```diff
@@ -0,0 +1,73 @@
...
+      searchUploadRequestEntriesInteractor = SearchUploadRequestEntriesInteractor();
+    });
+
+    test('search upload request entries should return success with results', () async {
+      final state = await searchUploadRequestEntriesInteractor.execute([uploadRequestEntry1, uploadRequestEntry2], SearchQuery('uploadRequestEntry1'));
+      state.fold(
+          (failure) => null,
+          (success) {
+        expect(success, isA<SearchUploadRequestEntriesSuccess>());
+        expect([sharedSpace1], (success as SearchUploadRequestEntriesSuccess).uploadRequestEntriesList);
```
huynguyennovem, line 52: LOL, it should be an array of `Upload Request Entry`. Not `sharedSpace1`

imGok:
> The message `No files uploaded yet` is blinking appear before the list of recipients is displayed.

I don't think it's a problem since we did that too in shared space + folder

domain/test/usecases/search_upload_request_entries/search_upload_request_entries_interactor_test.dart
imGok: ... you got me

huynguyennovem:
> > The message `No files uploaded yet` is blinking appear before the list of recipients is displayed.
> 
> I don't think it's a problem since we did that too in shared space + folder

I know that. But when the app is loading the recipient list, why the message is `No files uploaded yet`?

imGok:
> I know that. But when the app is loading the recipient list, why the message is `No files uploaded yet`?

Because the list of files is empty and we choose the background with list. it's the same as folder, shared space as I told you. It's better to do so than viewState

hoangdat:
> > I know that. But when the app is loading the recipient list, why the message is `No files uploaded yet`?
> 
> 
> 
> Because the list of files is empty and we choose the background with list. it's the same as folder, shared space as I told you. It's better to do so than viewState 

The loading and the empty state need to be separated

imGok:
For now, it displayed an empty list at start we need to fix the bug later (create emptylistindicator)

huynguyennovem (approved):

hoangdat (approved):

Issue #421, opened 2021-05-13
Redundant message when no result in search


hantt12:
### Description
In all screens (received share, myspace, shared space), if user searches and there is no result, the message is duplicate
In my opinion, we can remove "0 result" at the top

Tested in ios 0.7.0 (46)

![signal-2021-05-13-142742.jpeg](https://images.zenhubusercontent.com/5ee9def7c40b48447a4362e2/5e62ccda-17a7-4392-a7dc-4d9ac4815ea7)

huynguyennovem:
Confirm this bug is fixed in ver 2.0.0
https://github.com/linagora/linshare-mobile-flutter-app/commit/064ab96301359c34ed1c271e8dcca022041e3f12
![Screenshot_1633086286.png](https://images.zenhubusercontent.com/605175c4593ffb54122f0753/969aa8c3-1305-45b2-af35-f1164301c918)

PR #615, opened 2021-09-13
remove results when null in search view (#421)

imGok:
searchvewwe

### DIFF
--- a/lib/presentation/view/common/common_view.dart
+++ b/lib/presentation/view/common/common_view.dart
@@ -44,6 +44,10 @@ class CommonView {
   }
 
   Widget buildResultCountRow(BuildContext context, int count) {
+    if (count == 0) {
+      return SizedBox.shrink();
+    }
+
     return Container(
       color: AppColor.topBarBackgroundColor,
       height: 40.0,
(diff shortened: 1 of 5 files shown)


hoangdat:
How about the UI whenever no result matched? @imGok

lib/presentation/view/common/common_view.dart
```diff
@@ -44,6 +44,10 @@ class CommonView {
   }
 
   Widget buildResultCountRow(BuildContext context, int count) {
+    if (count == 0) {
```
huynguyennovem, line 47: You updated here but haven't used this function in the screens (my space, received share, shared space). 
If you use this, you don't need to check the length of state's list more :)

imGok: yh u are right i'll update other too

imGok: > How about the UI whenever no result matched? @imGok

image

imGok:
> How about the UI whenever no result matched? @imGok

image

imGok:
@hoangdat rdy to go

huynguyennovem (approved):
/lgtm

Issue #649, opened 2021-09-28
[Bug] File did not make offline is showing offline icon


hoangdat:
### Current behavior

1. Logged in with user 1
2. Select a workgroup `X`, which have member: (user1, user2, ...)
3. Select a file `F` on `X` then make available offline
4. Logout
5. Logged in with user2
6. Open the workgroup `X`
7. Check the file `F` 

### Acceptance criteria

The list of acceptance criteria (defined in the corresponding User Story or coming from your functional knowledge of the product)
These acceptance criteria must be met by the corrective implementation done by the developers.

### Context

Devices, OS

### Additional information

frequency, log files,...

huynguyennovem:
Can't re-procedure. Can you record a video?

PR #652, opened 2021-09-30
LA-649 Bug fixes offline mode when logout

huynguyennovem:
- Ticket: #649 
- Demo: 

https://user-images.githubusercontent.com/29337364/135487338-4185cca1-4ef7-4478-982d-aba5a92e7f9a.mp4


hoangdat:
@huynguyennovem please rebase it. One more question: how about if user close app right after logout (delete tasks is not completed)?

huynguyennovem:
> @huynguyennovem please rebase it. One more question: how about if user close app right after logout (delete tasks is not completed)?

This is hard to occur because we always `await` `logoutAction` stuff. After that finished, app navigates to login screen. 
Unless user actions are faster than phone chip speed or maybe the auto-test is set up for this case.

hoangdat (approved):

PR #697, opened 2021-11-05
LA-645 Add info button to the app bar of Workgroup

dab246:
https://user-images.githubusercontent.com/80730648/140453869-eb7789ed-b124-49f0-8ba5-3b703a76012f.mov


hoangdat:
How about when user in a child folder?

dab246:
> How about when user in a child folder?

button is still visible because it still belongs to that workgroup.

hoangdat (approved):

PR #729, opened 2021-11-10
[Drive] Support Drive function

hoangdat:
### separate context menu for Drive
- [x] Done

### open drive 
- [ ] all child workgroup of drive will be showed by API
```
@GET /webservice/rest/user/v4/shared_spaces?parent={id}&withRole=true
```

- [ ] support drive in destination picker
- [ ] temporary disable create workgroup inside drive


hoangdat:
duplicate with #730

PR #752, opened 2021-11-23
LA-704 Support account detail in offline mode

dab246:
https://user-images.githubusercontent.com/80730648/142981820-13a2384a-ffb2-4641-818b-51086e83d22f.mov


hoangdat:
how about if we haven't come to account details before?

hoangdat (approved):

PR #774, opened 2021-12-03
:hammer: Fix syntax error in bitrise.yaml

hoangdat:


hoangdat:
@tnstung I repair your fix :)

PR #793, opened 2022-01-04
LA-792 [BUG] Fix upload from outside can not upload to shared space

dab246:
https://user-images.githubusercontent.com/80730648/148010629-40632f22-ab5d-4a26-8889-8439880ca9b6.mov

### DIFF
--- a/lib/presentation/widget/home/home_viewmodel.dart
+++ b/lib/presentation/widget/home/home_viewmodel.dart
@@ -91,7 +91,6 @@ class HomeViewModel extends BaseViewModel {
       this._getBiometricSettingInteractor
   ) : super(store) {
     _getBiometricAuthenticationTimerSetting();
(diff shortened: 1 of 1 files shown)


hoangdat:
@dab246 can you desceribe sth about this bug?

dab246:
> can you describe something about this bug?

This error occurs when the application is completely closed. The reason is that getting data in functionality hasn't finished but the shared data listener is already running.

**Solution:**

- Start listening when the process of getting data from functionality is finished.

hoangdat:
> > can you describe something about this bug?
> 
> This error occurs when the application is completely closed. The reason is that getting data in functionality hasn't finished but the shared data listener is already running.
> 
> **Solution:**
> 
> * Start listening when the process of getting data from functionality is finished.

ohh, got it

hoangdat (approved):

PR #816, opened 2022-01-12
Add staging environment and build-time variable to config

hoangdat:
@dab246 please review


dab246 (commented):
IMO you shouldn't create too many repositories. Instead you should `define` output multiple `url` corresponding to each environment via the `SaaSType` variable.

hoangdat:
@dab246 please review again

dab246 (approved):
good !

PR #824, opened 2022-01-18
LA-820 Support WORK_SPACE

dab246:
https://user-images.githubusercontent.com/80730648/149924850-442ce0a6-060b-46ac-9499-88e752509251.mp4


hoangdat:
- How about upload inside/outside to WorkSpace?

dab246:
> * How about upload inside/outside to WorkSpace?

It works just like the `Drive` case

hoangdat (approved):

PR #833, opened 2022-01-21
Fix Deleting token

hoangdat:


data/lib/src/datasource_impl/authentication_oidc_datasource_impl.dart
```diff
@@ -74,9 +74,12 @@ class AuthenticationOIDCDataSourceImpl implements AuthenticationOIDCDataSource {
             redirectUrl,
             discoveryUrl: discoveryUrl,
             scopes: scopes,
-            preferEphemeralSession: preferEphemeralSessionIOS,
-            promptValues: promptValues,
-            allowInsecureConnections: allowInsecureConnections));
+            preferEphemeralSession: true,
+            promptValues: ['consent'],
+            responseMode: 'query'));
```
dab246, line 79: why hardcode? IMO should get the value from the config file.

hoangdat: This information is not related to config file, it is only depended on the client when connecting to server. No problem

dab246 (approved):

PR #855, opened 2022-03-24
Add crashlytics to LinShare project

hoangdat:
- [x] get crash log in local
- [ ] config to run with CI
- [ ] config to run with CD


hoangdat:
image

image

PR #908, opened 2022-09-28
Migrated to API key for release builds

tk-nguyen:
Since we got the key, we are migrating to it instead of using username and password


android/fastlane/Fastfile
```diff
@@ -25,15 +25,16 @@ platform :android do
...
-      track: "alpha"
+      track: track,
+      json_key_data: ENV["PLAY_STORE_CONFIG_JSON"]
     )[0].to_i
     build_name = last_git_tag.gsub("v", "")
     sh "flutter build apk --verbose --release --dart-define=CAPTCHA_TOKEN=$CAPTCHA_TOKEN --dart-define=SAAS_AVAILABLE=$SAAS_AVAILABLE --build-number=#{latest_build_number+1} --build-name=#{build_name}"
     upload_to_play_store(
       json_key_data: ENV["PLAY_STORE_CONFIG_JSON"],
-      track: "alpha",
+      track: track,
```
hoangdat, line 37: why not keep `alpha`

hoangdat: i see

hoangdat (approved):

PR #920, opened 2023-02-01
Fix chunk upload with ChunkedStreamReader to cancel underlying stream

hoangdat:
### Root cause:
- can not free resource when `file.openRead()` for each chunk

### Solution:
- using ChunkStreamReader to cancel the underlying stream


dab246 (commented):
Please check case: `_chunkedStreamReader?.cancel(); ` when `upload failed`

nqhhdev (approved):

PR #943, opened 2024-02-21
Fixed incorrect expiration date for files in mySpace

KhaledNjim:
### Description
[GitLab issue](https://ci.linagora.com/linagora/lgs/linshare/products/linshare-ui-user/-/issues/1220)

#### Root cause 
 - Wrong Key for expiration date in document_details  
 - The expiration date doesn't take into consideration undefined expiration date 
#### Solution
 - Called the right `expiration date`  key in document_details
 - Added handling for null `expiration_date`


hoangdat (commented):
srr, but it also include too much format change. hard for me.

PR #949, opened 2024-03-27
Bug fix/cannot add recipient to upload request

KhaledNjim:
## Description
[gitlab issue](https://ci.linagora.com/linagora/lgs/linshare/products/linshare-ui-user/-/issues/1206)

### Root cause
   The api for adding a recipient is not expecting a `uuid` field so its causing a bad request

### DIFF
--- a/data/lib/src/network/linshare_http_client.dart
+++ b/data/lib/src/network/linshare_http_client.dart
@@ -691,7 +691,7 @@ class LinShareHttpClient {
             .withPathParameter(uploadRequestGroupId.uuid)
             .withPathParameter(Endpoint.recipents)
             .generateEndpointPath(),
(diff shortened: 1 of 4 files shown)


data/lib/src/network/linshare_http_client.dart
```diff
@@ -691,7 +691,7 @@ class LinShareHttpClient {
             .withPathParameter(uploadRequestGroupId.uuid)
             .withPathParameter(Endpoint.recipents)
             .generateEndpointPath(),
-        data: recipients.map((data) => GenericUserDto(data.mail, lastName: data.lastName, firstName: data.firstName).toJson()).toList().toString());
```
hoangdat, line 694: you can ignore fields in `toJson` by annotation, no need to create another extensions

hoangdat (approved):

PR #953, opened 2024-04-16
Fixed donwloads in sharedSpace

KhaledNjim:
### Description
[GitLab issue](https://ci.linagora.com/linagora/lgs/linshare/products/linshare-ui-user/-/issues/1201)
#### Root cause 
  Flutter_downloader requires `saveInPublicStorage` flag to be set to true for devices with android version greater than android Q(10) to be able to save in downloads folder
#### Solution
 Set the `saveInPublicStorage` flag to `true` in  the  Flutter_Downloader call in `Download_document method`


hoangdat (approved):

dieptran88:
fixed on 3.10.0 version

PR #956, opened 2024-05-17
LA-1296 changed key value for oidc-Jwt

KhaledNjim:
## Description 
[gitlab issue](https://ci.linagora.com/linagora/lgs/linshare/products/linshare-ui-user/-/issues/1296)
  
when trying to login using sso with oidc-jwt the authentication fails

## Root cause
 The `oidc_parser` is comparing to a wrong value for oidc token type `JWT` instead of  `Oidc-Jwt` so it's always defaulting to using `Oidc-opaque`

## Solution 
Changed to the right value

### DIFF
--- a/data/lib/src/util/oidc_parser.dart
+++ b/data/lib/src/util/oidc_parser.dart
@@ -25,7 +25,7 @@ class OIDCParser {
       final responseTypeMatch = regExpResponseType.allMatches(matchString).first.group(3) ?? '';
 
       String oidcTokenType = DomainConstant.opaqueOidc;
-      if (oidcTypeMatch == 'Oidc-Jwt') {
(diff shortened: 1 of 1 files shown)


data/lib/src/util/oidc_parser.dart
```diff
@@ -25,7 +25,7 @@ class OIDCParser {
       final responseTypeMatch = regExpResponseType.allMatches(matchString).first.group(3) ?? '';
 
       String oidcTokenType = DomainConstant.opaqueOidc;
-      if (oidcTypeMatch == 'JWT') {
+      if (oidcTypeMatch == 'Oidc-Jwt') {
```
hoangdat, line 28: please add constant for this.

PR #958, opened 2024-05-20
LA-1295 Fixed missing password protected checkbox

KhaledNjim:
## Description 
[gitlab issue](https://ci.linagora.com/linagora/lgs/linshare/products/linshare-ui-user/-/issues/1295)

### Root cause
  Wrong attribute is used to determine if the checkbox should be displayed or not

### Solution
Changed the attribute to `protectPasswordSetting` instead of  `passwordProtected`

### DIFF
--- a/lib/presentation/widget/upload_request_creation/upload_request_creation_widget.dart
+++ b/lib/presentation/widget/upload_request_creation/upload_request_creation_widget.dart
@@ -169,7 +169,7 @@ class _UploadRequestCreationWidgetState extends State<UploadRequestCreationWidge
                         ..addOnSizeTypeNotifier(_model.totalFileSizeTypeNotifier))
                       .build()
                     : null)
-                ..addPasswordProtectedInput(creationState.uploadRequestCreation?.passwordProtected != null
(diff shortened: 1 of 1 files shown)


lib/presentation/widget/upload_request_creation/upload_request_creation_widget.dart
```diff
@@ -169,7 +169,7 @@ class _UploadRequestCreationWidgetState extends State<UploadRequestCreationWidge
                         ..addOnSizeTypeNotifier(_model.totalFileSizeTypeNotifier))
                       .build()
                     : null)
-                ..addPasswordProtectedInput(creationState.uploadRequestCreation?.passwordProtected != null
+                ..addPasswordProtectedInput(creationState.uploadRequestCreation?.protectPasswordSetting != null
```
hoangdat, line 172: Please careful on it, maybe conflict with the older version of back-end

KhaledNjim: i tested with old versions and it works as intented

hoangdat (approved):

PR #959, opened 2024-05-20
LA-1294 Fixed error in logs in workgoup activities

KhaledNjim:
## Desciption
[gitlab issue](https://ci.linagora.com/linagora/lgs/linshare/products/linshare-ui-user/-/issues/1294)

### Root cause 

  The activity logs causing the error have the type `WORK_GROUP`  instead of  `WORKGROUP`  which leads to a problem while parsing
image
### Solution  
 Changed  `WORK

GROUP` to `WORK_GROUP`
Screenshot_1

### DIFF
--- a/data/lib/src/network/linshare_http_client.dart
+++ b/data/lib/src/network/linshare_http_client.dart
@@ -200,7 +200,7 @@ class LinShareHttpClient {
   }
 
   AuditLogEntryUserDto? _convertToAuditLogEntryNodeChild(Map<String, dynamic> nodeChildJson) {
-    if (nodeChildJson['type'] == AuditLogEntryType.WORKGROUP.value) {
+    if (nodeChildJson['type'] == AuditLogEntryType.WORK_GROUP.value) {
       return SharedSpaceNodeAuditLogEntryDto.fromJson(nodeChildJson);
     } else if (nodeChildJson['type'] == AuditLogEntryType.WORKGROUP_MEMBER.value) {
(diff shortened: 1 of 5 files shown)


data/lib/src/network/linshare_http_client.dart
```diff
@@ -200,7 +200,7 @@ class LinShareHttpClient {
   }
 
   AuditLogEntryUserDto? _convertToAuditLogEntryNodeChild(Map<String, dynamic> nodeChildJson) {
-    if (nodeChildJson['type'] == AuditLogEntryType.WORKGROUP.value) {
+    if (nodeChildJson['type'] == AuditLogEntryType.WORK_GROUP.value) {
```
hoangdat, line 203: please very careful, make sure it still has good interoperability

hoangdat: can you add the link?

KhaledNjim: the issue i found was with the version 4 
https://user.linshare-4-2-on-commit.integration-linshare.org

hoangdat: What you want to share with this link? `WORKGROUP` & `WORK_GROUP` also work?

KhaledNjim: for newer api v5 and v6 its `WORK_GROUP` and for older api v4 its `WORKGROUP`, other attributes are the same

hoangdat (approved):

PR #970, opened 2024-11-20
LA-1355 fix upload request total max size

KhaledNjim:
## Root cause
 - The current conditions compare values that might be in different units
  - The error message contains a hardcoded value of 100GB, which might not reflect the actual limit.
 ## Solution
 - Converted both the input value and limit to bytes before comparing.
 - Added parameters for the l10n string to display the actual limit.
 ## Demo
 

https://github.com/user-attachments/assets/fd3e6991-f393-4b5a-8764-524e0a810cae

### DIFF
--- a/lib/l10n/intl_en.arb
+++ b/lib/l10n/intl_en.arb
@@ -1639,11 +1639,17 @@
     "placeholders_order": [],
     "placeholders": {}
   },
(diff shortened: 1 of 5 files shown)


hoangdat (approved):

hoangdat:
please rebase

KhaledNjim:
> please rebase

Done

hoangdat (approved):

Issue #971, opened 2024-11-29
Play store reject LinShare app 


hoangdat:
### Description
- `APP HAS A PRIVACY POLICY​ IN PLAY DEVELOPER CONSOLE BUT IT IS NOT ADEQUATE`
<img width="1177" alt="image" src="https://github.com/user-attachments/assets/557b71bb-503a-4a86-b9c2-f4f799556aba">

- `User Data policy: APK REQUIRES PROMINENT DISCLOSURE​`
<img width="937" alt="image" src="https://github.com/user-attachments/assets/ce869f01-76de-4c7f-b481-e392d7488c47">

- `App must target Android 14 (API level 34) or higher`

### Expected result

- [ ] can pass the evaluation 

Cc: @wboudiche @KhaledNjim

PR #972, opened 2024-12-19
LA-971 added explanation for contact permission

KhaledNjim:
## Issue
 - #971
## Solution
  - added dialog to explain the use of contacts permission before requesting it
 ## Demo 
 

https://github.com/user-attachments/assets/b76b944e-840a-432e-8dea-7f9123a8ab6e

### DIFF
--- a/lib/presentation/view/dialog/permission_dialog.dart
+++ b/lib/presentation/view/dialog/permission_dialog.dart
@@ -2,7 +2,7 @@ import 'package:flutter/material.dart';
 import 'package:linshare_flutter_app/presentation/localizations/app_localizations.dart';
 
 class PermissionDialog {
-  static Future<bool?> showPermissionDialog(
+  static Future<bool?> showPermissionExplanationDialog(
       BuildContext context, Widget title, String content) async {
     return showDialog<bool>(
         barrierDismissible: false,
--- a/lib/presentation/view/dialog/permission_dialog.dart
+++ b/lib/presentation/view/dialog/permission_dialog.dart
@@ -1,0 +1,25 @@
+import 'package:flutter/material.dart';
+import 'package:linshare_flutter_app/presentation/localizations/app_localizations.dart';
+
+class PermissionDialog {
+  static Future<bool?> showPermissionDialog(
+      BuildContext context, Widget title, String content) async {
+    return showDialog<bool>(
+        barrierDismissible: false,
+        context: context,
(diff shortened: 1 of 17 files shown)


hoangdat:
hi @KhaledNjim, we should handle it for permission in image/video or files also

KhaledNjim:
> hi @KhaledNjim, we should handle it for permission in image/video or files also

Updated, added explanation before requesting permission for image/video microphone and phone state.
  for files permission is not requested
  

https://github.com/user-attachments/assets/0fa79fdc-7fd2-4cc4-bad8-254bcc26df7b

hoangdat:
- [ ] you should mention clearly which function to make user understand about your permission request for what. I suggest: `To continue to recording video ....`.

hoangdat:
permission explanation still display after user accepted permission 

### DoD:
- [ ] should not display after user granted permission


https://github.com/user-attachments/assets/d65856a1-79f5-42cf-a4f8-66dc24c11ce0

KhaledNjim:
> * [ ]  you should add specific explanation for these 2 options also (don't forget if it uses in other spaces(shared space, received share)
> 
> Screenshot_2024-12-30-05-34-42-670_com linagora android linshare

In those cases, no permissions are requested

KhaledNjim:
> * [ ]  you should mention clearly which function to make user understand about your permission request for what. I suggest: `To continue to recording video ....`.
> 
> 

Updated 

image

hoangdat:
> In those cases, no permissions are requested

No, these still need request permission, so please add the explanation for these too.

IMG_0187

lib/presentation/view/dialog/permission_dialog.dart
```diff
@@ -0,0 +1,25 @@
+import 'package:flutter/material.dart';
+import 'package:linshare_flutter_app/presentation/localizations/app_localizations.dart';
+
+class PermissionDialog {
+  static Future<bool?> showPermissionDialog(
```
hoangdat, line 5:
```suggestion
  static Future<bool?> showPermissionExplanationDialog(
```

KhaledNjim:
> > In those cases, no permissions are requested
> 
> No, these still need request permission, so please add the explanation for these too.
> 
> IMG_0187

Updated, this permission is only requested on api 32 and lower
image

hoangdat (approved):

